Issue #9, opened 2024-11-07
Add caching for faster downloads of torch


AndreSlavescu:

msaroufim:
This was resolved thanks to modal

PR #10, opened 2024-11-07
dependency caching

AndreSlavescu:
re: 

https://github.com/gpu-mode/discord-cluster-manager/issues/9

Results after run locally (macbook pro m1)

```
==================================================
                BENCHMARK RESULTS                 
==================================================
Test Type                     Duration            
--------------------------------------------------
With Cache                    26.76s              
Without Cache                 68.07s              
With UV                       9.28s               
Cache Speedup                 41.31s              
UV vs Cache Speedup           17.48s              
==================================================

Test Output:
--------------------------------------------------
Matrix multiplication result:
 [[19 22]
 [43 50]]```

### DIFF
--- a/.github/workflows/train_workflow.yml
+++ b/.github/workflows/train_workflow.yml
@@ -36,7 +36,7 @@ jobs:
         run: |
           uv venv
           source .venv/bin/activate
-          uv pip install --no-deps numpy torch
+          uv pip install --no-deps numpy torch typing_extensions
           
       - name: Create and run training script
         run: |
--- a/.github/workflows/train_workflow.yml
+++ b/.github/workflows/train_workflow.yml
@@ -32,7 +32,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip install numpy torch
-          pip install -r requirements.txt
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: .github/workflows/train_workflow.yml [hunk 1]
```diff
@@ -5,17 +5,35 @@ on:
...
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
       - name: Install dependencies
         run: |
-          pip install numpy
-          # pip install torch - need to find a way to cache this otherwise it will take a long time to install
-
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
```
msaroufim, line 35: we don't need most these requirements to run the file, we mostly need them to run the bot which is not yet properly tested in CI

test_workflow.py
```diff
@@ -0,0 +1,58 @@
+import subprocess
```
msaroufim, line 1: the github action should report how long it takes to run which will be more indicative of actual performance. see my caching prs i noticed caching would be slower than without

PR #11, opened 2024-11-07
torch example

AndreSlavescu:
re:

#9


torch/train.py
```diff
@@ -0,0 +1,8 @@
+import torch
```
msaroufim, line 1: this is small enough to combo in 1 PR

PR #15, opened 2024-11-09
Staging name

AndreSlavescu:


msaroufim:
About to head to bed, @S1ro1 if you can also test this locally this should be good to merge

S1ro1:
We should try to cleanup the bot asap, passing variables around when the bot is created like this gets very painful very soon, for now merging this with a hack.

PR #18, opened 2024-11-12
Add Numpy / Torch / Triton differentiation

alexzhang13:
Install numpy / Triton / torch based on grepping the train.py. If it's too slow, we can change this in the future.


msaroufim (commented):
interesting trick although it would break if someone does something like `from torch import nn`

I think to make this work properly we can do clever ast stuff like https://gist.github.com/msaroufim/2df6d47a0164b3524eefb97aa23f6696

msaroufim (approved):

PR #26, opened 2024-11-19
S1ro1/run cuda

S1ro1:
This PR enables running CUDA kernels same way as python scripts are currently supported by submitting a file named `train.cu`.


S1ro1:
cc: @alexzhang13 @msaroufim let's move the discussion around this here.

PR #36, opened 2024-11-24
add /verifyruns cog

b9r5:
## Description

Adds a `/verifyrun` command that does the same work as the smoke test but is easier to use.

image

## Checklist

- [x] Run the smoke test on your own server.
    ```bash
    python discord-bot.py
    ```
  - Start a training run by with the slash command `/run`.
    You may need to exercise some judgement about the script and GPU type.
  - Wait for the training run to complete.
  - Copy the URL for the thread started by the cluster bot in response to
    your `/run` message ("Cluster Bot started a thread: ..."):
      - Click on the 3 dots (`...`) to the cluster bot's message.
      - Select *Copy Message Link*.
  - Using the copied URL, run the smoke test:
    ```bash
    python discord-bot-smoke-test.py copied_url
    ```
  - Verify that the smoke test script responds with:
    ```
    All tests passed!
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/cogs/github_cog.py
+++ b/src/discord-cluster-manager/cogs/github_cog.py
@@ -35,6 +35,7 @@ class GitHubCog(commands.Cog):
         interaction: discord.Interaction,
         script: discord.Attachment,
         gpu_type: app_commands.Choice[str],
+        use_followup: bool = False
     ):
         if not script.filename.endswith(".py") and not script.filename.endswith(".cu"):
             await interaction.response.send_message(
@@ -43,10 +44,15 @@ class GitHubCog(commands.Cog):
             return
 
         thread = await self.bot.create_thread(interaction, gpu_type.name, "GitHub Job")
+        message = f"Created thread {thread.mention} for your GitHub job"
 
-        await interaction.response.send_message(
-            f"Created thread {thread.mention} for your GitHub job"
-        )
+        if use_followup:
+            if not interaction.response.is_done():
+                await interaction.response.defer()
+            await interaction.followup.send(message)
+        else:
+            await interaction.response.send_message(message)
+
         await thread.send(f"Processing `{script.filename}` with {gpu_type.name}...")
 
         try:
--- a/src/discord-cluster-manager/cogs/github_cog.py
+++ b/src/discord-cluster-manager/cogs/github_cog.py
@@ -47,8 +47,6 @@ class GitHubCog(commands.Cog):
         message = f"Created thread {thread.mention} for your GitHub job"
 
         if use_followup:
-            if not interaction.response.is_done():
-                await interaction.response.defer()
             await interaction.followup.send(message)
         else:
             await interaction.response.send_message(message)
--- a/src/discord-cluster-manager/cogs/github_cog.py
+++ b/src/discord-cluster-manager/cogs/github_cog.py
@@ -36,12 +36,12 @@ class GitHubCog(commands.Cog):
         script: discord.Attachment,
         gpu_type: app_commands.Choice[str],
         use_followup: bool = False
-    ):
+    ) -> discord.Thread:
         if not script.filename.endswith(".py") and not script.filename.endswith(".cu"):
             await interaction.response.send_message(
                 "Please provide a Python (.py) or CUDA (.cu) file"
             )
-            return
+            return None
 
         thread = await self.bot.create_thread(interaction, gpu_type.name, "GitHub Job")
         message = f"Created thread {thread.mention} for your GitHub job"
@@ -85,6 +85,9 @@ class GitHubCog(commands.Cog):
         except Exception as e:
             logger.error(f"Error processing request: {str(e)}", exc_info=True)
             await thread.send(f"Error processing request: {str(e)}")
+
+        finally:
+            return thread
 
     async def trigger_github_action(self, script_content, filename, gpu_type):
         logger.info(f"Attempting to trigger GitHub action for {gpu_type.name} GPU")
--- a/src/discord-cluster-manager/cogs/verify_run_cog.py
+++ b/src/discord-cluster-manager/cogs/verify_run_cog.py
@@ -3,8 +3,19 @@ from discord import app_commands
 from discord.ext import commands
 import re
 from utils import setup_logging
+from unittest.mock import AsyncMock
 
 logger = setup_logging()
+
+def create_mock_attachment(filename: str, content: str):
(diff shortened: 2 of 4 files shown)


src/discord-cluster-manager/cogs/verify_run_cog.py
```diff
@@ -0,0 +1,103 @@
...
+logger = setup_logging()
+
+class VerifyRunCog(commands.Cog):
+    """
+    A Discord cog for verifying the success of trainingruns.
+
+    This cog provides functionality to verify that either a GitHub Actions or
+    Modal run completed successfully by checking for specific message patterns
+    in a Discord thread. It supports verification of two types of runs:
+    1. GitHub Actions runs - Identified by "GitHub Action triggered!" message
```
msaroufim, line 16: So this seems to test a trigger but we've had issues where this wouldn't have helped like for example when we had timeout issues with both the NVIDIA and AMD runner

msaroufim (commented):
There is some time we could save in having a single command verify all of modal, nvidia and amd one after the other and letting us know how many of those 3 succeeded. Also as far as the UX goes, why not have verify actually trigger the bot 3 times instead of having a human manually verify in a thread

b9r5:
OK, thanks, Mark. I'll switch `/verifyrun` to start all 3 types of runs. And I'll make sure I'm not waiting for "GitHub Action triggered!" messages that may never come.

I switched this PR to draft for now because I'm in the middle of re-doing this.

b9r5:
Latest changes launch 3 run jobs.

Screenshot below. I find the bot's messages like "Created thread ⁠GitHub Job (AMD) - 2024-11-29 1… for your GitHub job" a little distracting, given that the threads also show up as separate lines in the UI. Should we remove those "created thread" messages?

@msaroufim, I don't know if this addresses your comment from above ("we've had issues where this wouldn't have helped"). If a GitHub workflow times out, then I think the GitHub cog will just return, and the thread won't have the happy path messages in it, so verification will fail. So I _think_ it addresses your comment, but I'm not fully sure.

Screenshot from 2024-11-29 11-58-22

msaroufim:
> Screenshot below. I find the bot's messages like "Created thread ⁠GitHub Job (AMD) - 2024-11-29 1… for your GitHub job" a little distracting, given that the threads also show up as separate lines in the UI. Should we remove those "created thread" messages?

Yeah should remove those in another PR

> @msaroufim, I don't know if this addresses your comment from above ("we've had issues where this wouldn't have helped"). If a GitHub workflow times out, then I think the GitHub cog will just return, and the thread won't have the happy path messages in it, so verification will fail. So I think it addresses your comment, but I'm not fully sure.

Yeah i recently implemented timeouts so they should solve that specific problem

b9r5:
Mark, do you think this needs a final message at the end that says "✅ All tests passed"?

msaroufim:
Yeah that sounds nice

b9r5:
I added the final message.

But I'm getting a bad request response intermittently: 
```
2024-12-04 07:23:41 - utils - ERROR - Error starting verification runs: 400 Bad Request (error code: 40060): Interaction has already been acknowledged.
Traceback (most recent call last):
  File "/home/benhorowitz/Code/Discord-cluster-manager/src/discord-cluster-manager/cogs/verify_run_cog.py", line 137, in verify_runs
    results = await asyncio.gather(
              ^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        self.verify_modal_run(modal_cog, interaction))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benhorowitz/Code/Discord-cluster-manager/src/discord-cluster-manager/cogs/verify_run_cog.py", line 89, in verify_modal_run
    modal_thread = await modal_command.callback(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        modal_cog, interaction, script_file, t4, use_followup=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benhorowitz/Code/Discord-cluster-manager/src/discord-cluster-manager/cogs/modal_cog.py", line 45, in run_modal
    await interaction.response.defer()
  File "/home/benhorowitz/Code/Discord-cluster-manager/.venv/lib/python3.13/site-packages/discord/interactions.py", line 709, in defer
    await adapter.create_interaction_response(
    ...<6 lines>...
    )
  File "/home/benhorowitz/Code/Discord-cluster-manager/.venv/lib/python3.13/site-packages/discord/webhook/async_.py", line 223, in request
    raise HTTPException(response, data)
discord.errors.HTTPException: 400 Bad Request (error code: 40060): Interaction has already been acknowledged.
```

I think this is due to a race condition in this type of logic:
```
            if not interaction.response.is_done():
                await interaction.response.defer()
```
What I believe is happening is that two asyncs execute the first line, both tests pass, and then two try to execute the second line. I'll fix this up.

b9r5:
image

src/discord-cluster-manager/cogs/github_cog.py
```diff
@@ -35,18 +35,22 @@ async def run_github(
...
                 "Please provide a Python (.py) or CUDA (.cu) file"
             )
-            return
+            return None
 
         thread = await self.bot.create_thread(interaction, gpu_type.name, "GitHub Job")
+        message = f"Created thread {thread.mention} for your GitHub job"
+
+        if use_followup:
+            await interaction.followup.send(message)
```
msaroufim, line 50: you might have answered this already but remind me what does followup mean?

msaroufim:
Minor question on the bot.py changes, also have a merge conflict otherwise this is easy to merge

src/discord-cluster-manager/cogs/github_cog.py
b9r5: So interactions can be replied to with `interaction.response.send_message(...)`. But in this case that's the last message that the bot can send for the interaction. And the other option is to use `interaction.followup.send(...)`, which lets multiple messages be sent in response to the interaction. 

As to why this distinction exists, I believe the reason is that interactions need timely responses -- within 3 seconds. So if the response is going to take longer than that, as it does with training runs, then followups have to be used.

By the way, VerifyRunCog has to set up the interaction for followups by calling `interaction.response.defer()`.

Hope that explains things. LMK if you have more questions!

msaroufim (approved):

PR #42, opened 2024-12-07
leaderboard schema

b9r5:
## Description

We want to create tables where we can store leaderboard data. 

## Checklist

- [x] Run the smoke test on your own server.
    ```bash
    python discord-bot.py
    ```
  - Start a training run by with the slash command `/run`.
    You may need to exercise some judgement about the script and GPU type.
  - Wait for the training run to complete.
  - Copy the URL for the thread started by the cluster bot in response to
    your `/run` message ("Cluster Bot started a thread: ..."):
      - Click on the 3 dots (`...`) to the cluster bot's message.
      - Select *Copy Message Link*.
  - Using the copied URL, run the smoke test:
    ```bash
    python discord-bot-smoke-test.py copied_url
    ```
  - Verify that the smoke test script responds with:
    ```
    All tests passed!
    ```
  README.md.


b9r5:
- I created a leaderboard schema because we might want other schemas, and we might want to use the public schema for other things.
- problem table:
  - I renamed the leaderboard table to problem because it seemed like the table is about problems.
  - I changed the type of problem.deadline to TIMESTAMP WITH TIME ZONE.
  - Note: this means that the timestamp is stored in UTC internally. It doesn’t mean that the timestamp is stored with a non-zero numeric offset or time zone like America/Los_Angeles.
  - I generally prefer including WITH TIME ZONE because it makes it clear that the time is stored in UTC.
  - It seemed like problem.reference_code should be UNIQUE, so I added that constraint.
  - I’m not sure about the reference_code column. I’m happy to keep it, but not sure how it would be used. Is this a short code that a person submitting a problem would create? Or is it a code that we would create for them? If we create it for them, could we use the primary key instead?
- submission table:
  - I unpluralized leaderboard.submissions because non-plural names seem like a good convention.
  - I renamed submission.submission_id to submission.id because having a primary key named id for all tables seems like a good idea.
  - I made submission.problem_id be a foreign key to problem.id. This is really a matter of taste, but I tend to like foreign keys that reference the primary key of the target table.
  - I renamed submission.submission_name to submission.name and changed its type to TEXT. I’m going back and forth on TEXT vs. VARCHAR(255) and basically can’t decide but am leaning toward TEXT. My thought is that we could check length limits in code if we need to, rather than enforcing a limit at the column level, and then we could be flexible if we need to relax whatever length limit we choose.
  - I renamed submission.submission_score to score. I made it NUMERIC rather than double because generally I think people like to see decimal numbers rather than binary floating point numbers.
  - I’m not sure about submission.score being not null. Wouldn’t it be null when first submitted and then set to some value later? Or would it be 0 when first submitted and then set later?
  - I think we’re going to need an index on problem_id. Postgres doesn’t add indexes for foreign keys automatically. (TODO)
- runinfo table:
  - I gave leaderboard.runinfo a primary key. It’s good to be able to identify rows by something, so I shy away from tables without primary keys.
  - Would we want stdout or ncu_output to be NOT NULL? Not sure, just wanted to ask.
  - I think we’re going to need an index on submission_id. (TODO)

b9r5:
## Expand/migrate/contract

I anticipate that initially we’ll want to iterate on the schema, and during this time it’ll be acceptable to drop tables and recreate them. After that initial period of iteration,  I am hoping that we can follow an expand/migrate/contract pattern to allow database migrations without downtime. This would involve doing the following:

- When you want to make a database change, determine if it is expansive or contractive.
  - Expansive changes have no possibility of breaking a running application. Examples are: adding a new nullable column, adding a non-null column with a default value, adding an index, adding a table, etc.
  - Contractive changes are those that could break a running application. Examples are: dropping a table, dropping a column, adding a not null constraint to a column, adding a unique index, etc.
- After an expansive phase, data gets migrated to the newly added elements. Code also begins using the newly added elements. This is the migration step.
- Finally, when all code is no longer using elements that are obsolete, these can be removed. Or, if adding a unique or not null constraint, after checking that the data satisfies the constraint, then the constraint can be safely added.

Expand/migrate/contract takes some discipline to use, but is also not too onerous once one is used to it, and it will help us avoid service interruptions.

## Applying schema changes

For applying schema changes, what I am thinking is the following. This is essentially what some migration frameworks do. I think it will save us some trouble later. Please let me know what you think.

- Create a table public.migration that has these columns:
  - filename TEXT NOT NULL PRIMARY KEY
  - sha256 BYTEA NOT NULL
  - applied_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP
- Create numbered sql files in a new directory src/discord-cluster-manager/db.
  - Files in this directory would be numbered starting with 0000.sql, which would  be the initial schema above.
- Write a script that does the following:
  - Back up the database.
    - It looks like Heroku is reselling Amazon RDS, so we might be able to use a snapshot.
    - Otherwise we’d have to work out where to store the backup. We could probably start with a Google drive or somewhere similar.
  - List at the numbered sql files in the db directory.
  - Sort them by their number.
  - Starting at the lowest number and proceeding to the highest:
    - Calculate the sha256 of the file.
      - If there is an entry in public.migration having the same filename:
        - If the sha256 is different, print an error message and exit (go no further – essentially a panic).
        - If the same, continue to the next file.
      - Else, in a single transaction:
        - Insert into public.migration
        - Execute the SQL in the file
        - Commit

b9r5:
@alexzhang13 wrote:

>It seemed like problem.reference_code should be UNIQUE

Not sure if others have opinions on this, but it’s technically possible for two problems to have the same reference code — they’d ofc just have the same functionality, but there’d be no need to check for uniqueness of code (idk how slow this would be too)

b9r5:
@alexzhang13 wrote:

> Is this a short code that a person submitting a problem would create?

This will be the code the problem creator will submit. So it should include the input generator and the reference solution. On our end, we’ll probably write a script that takes this code and verifies both correctness and runtime of a user submitted script against this reference code.

b9r5:
@alexzhang13 wrote:

>Or would it be 0 when first submitted and then set later?

The way I’ve set it up now it writes the entire submission to the table AFTER evaluating the GH runners. Open to having this be different, but also I agree about removing the not NULL condition

b9r5:
@alexzhang13 wrote:

>Would we want stdout or ncu_output to be NOT NULL? Not sure, just wanted to ask.

I guess if they’re empty it should be NULL, but I have no preference here

b9r5:
> @alexzhang13 wrote:
> 
> > Is this a short code that a person submitting a problem would create?
> 
> This will be the code the problem creator will submit. So it should include the input generator and the reference solution. On our end, we’ll probably write a script that takes this code and verifies both correctness and runtime of a user submitted script against this reference code.

I was super confused about this. I thought the reference code was a short name that was used to refer to the problem, like `softmax_fp8`. But it's the text of the program used to verify solutions to the problem. 

I removed the unique constraint on reference_code.

b9r5:
> @alexzhang13 wrote:
> 
> > Or would it be 0 when first submitted and then set later?
> 
> The way I’ve set it up now it writes the entire submission to the table AFTER evaluating the GH runners. Open to having this be different, but also I agree about removing the not NULL condition

It's always easier to remove a constraint than to add it. So if you're writing to the table after doing the run, then it's easier for keep the not null constraint for now, and remove it later if we need to.

b9r5:
> @alexzhang13 wrote:
> 
> > Would we want stdout or ncu_output to be NOT NULL? Not sure, just wanted to ask.
> 
> I guess if they’re empty it should be NULL, but I have no preference here

I think if they're empty they would be empty strings, and if we don't capture them they would be null. Are we always planning on capturing them?

I'm leaning towards keeping these nullable.

b9r5:
I'm thinking we could use yoyo-migrations for applying schema changes. I've never used it before, but it seems to do everything I would want in a migration tool.

alexzhang13 (approved):
Schema looks good and aligned with what I was thinking as well, other changes all are pretty standard and don't look problematic.

I'm going to assume yoyo changes are correct and running here.

README.md
```diff
@@ -43,13 +43,63 @@ Finally, generate an invite link for the bot and enter it into any browser.
 > [!NOTE]
 > Bot permissions involving threads/mentions/messages should suffice, but you can naively give it `Administrator` since it's just a test bot in your own testing Discord server.  
 
+### Database
+
+The leaderboard persists information in a Postgres database. To develop locally, set Postgres up on your machine. Then create a database with:
+
+```
+CREATE DATABASE clusterdev;
```
msaroufim, line 51: where do i type this in? just shell?

README.md
```diff
@@ -43,13 +43,63 @@ Finally, generate an invite link for the bot and enter it into any browser.
...
+    <summary>Click here for a transcript of a yoyo apply session</summary>
+
+    $ yoyo apply . -d postgresql://user:password@localhost/clusterdev
+
+    [20241208_01_p3yuR-initial-leaderboard-schema]
+    Shall I apply this migration? [Ynvdaqjk?]: y
+
+    Selected 1 migration:
+      [20241208_01_p3yuR-initial-leaderboard-schema]
+    Apply this migration to postgresql://user:password@localhost/clusterdev [Yn]: y
```
msaroufim, line 71: sorry n00b q: wdym by migration here?

msaroufim (approved):
some basic questions, but feel free to merge when ready

README.md
b9r5: In the Postgres shell. I'll edit the readme to include more details.

README.md
b9r5: So the term "migrations" refers to a type of version control for databases. The general idea is to specify changes to a database schema with a sequence of files. These files contain instructions, often in the form of SQL statements, that change database structure. As the application evolves, new migration files get added, so that the database structure changes along with the application. That that the files are also stored in version control provides a historical view of how the database changed in structure over time. 

A very typical use case is to add a column as part of a migration, and then to populate the column using existing data.

There is a whole family of tools for migrations in different environments. For example, Ruby on Rails has ActiveRecord migrations, Java has Liquibase and Flyway, Python has Alembic and Yoyo, and so on. The common flavor of these tools is that they allow CI/CD systems (or, in the old days, devops) to apply migrations in the same order across different environments (local dev, staging, prod, etc.).

src/discord-cluster-manager/migrations/20241208_01_p3yuR-initial-leaderboard-schema.py
```diff
@@ -0,0 +1,46 @@
+"""
```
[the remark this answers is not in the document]
b9r5: >how come this is dated?

I believe it's dated just as a kind of sequence number. I actually prefer numbering migrations in increasing order without gaps. Some of the yoyo documentation has that style; however `yoyo new` uses dates, and I thought it would be less confusing to stick with the dates rather than change the generated file name.

>is the idea that we version the schema using these python files?

Yes, that's exactly the idea. Versioning the schema makes the process of rolling the version forward less complex, and it provides a useful view of recent changes to the schema.

PR #52, opened 2024-12-13
QoL changes: Add GPU types view, remove dtype and shapes info. No DB or functionality.

alexzhang13:
A bunch of quality of life changes, as well as a change to GPU types.

Adds:
* Checkers and ephemeral info when leaderboard creation is invalid (duplicate key, invalid dates, etc.) in try-catch.
* New option and view to select GPU during problem creation.
image

Removes:
* `dtype` and `shape` information in `/leaderboard submit ...`.
* `dtype` and `shape` information in `/leaderboard create ...`.

Leftover TODOs:
* [ ] (urgent ⚠️!) Add DB changes to reflect new GPUs. @b9r5 let's talk about this more, but I think we should just have either score point to a DB with each valid GPU type (eh), or add a score for each possible GPU in the DB.
* [ ] Add leaderboard delete (dangerous! we should also have special UI for this so users are sure)
* [ ] Update `/leaderboard list` to also include all available GPU types for the particular problem.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/cogs/leaderboard_cog.py
+++ b/src/discord-cluster-manager/cogs/leaderboard_cog.py
@@ -1,6 +1,8 @@
 import discord
+from discord import ui, SelectOption, Interaction
 from discord import app_commands
 from discord.ext import commands
+from discord.app_commands.errors import CommandInvokeError
 from datetime import datetime
 
 from typing import TYPE_CHECKING
@@ -23,8 +25,6 @@ class LeaderboardSubmitCog(app_commands.Group):
     @app_commands.describe(
         leaderboard_name="Name of the competition / kernel to optimize",
         script="The Python / CUDA script file to run",
-        dtype="dtype (e.g. FP32, BF16, FP4) that the input and output expects.",
-        shape="Data input shape as a tuple",
     )
     # TODO: Modularize this so all the write functionality is in here. Haven't figured
     # a good way to do this yet.
@@ -33,8 +33,6 @@ class LeaderboardSubmitCog(app_commands.Group):
         interaction: discord.Interaction,
         leaderboard_name: str,
         script: discord.Attachment,
-        dtype: app_commands.Choice[str] = None,
-        shape: app_commands.Choice[str] = None,
     ):
         pass
 
@@ -53,8 +51,6 @@ class LeaderboardSubmitCog(app_commands.Group):
         leaderboard_name: str,
         script: discord.Attachment,
         gpu_type: app_commands.Choice[str],
-        dtype: app_commands.Choice[str] = "fp32",
-        shape: app_commands.Choice[str] = None,
     ):
         try:
             # Read the template file
@@ -110,8 +106,6 @@ class LeaderboardSubmitCog(app_commands.Group):
         leaderboard_name: str,
         script: discord.Attachment,
         gpu_type: app_commands.Choice[str],
-        dtype: app_commands.Choice[str] = "fp32",
-        shape: app_commands.Choice[str] = None,
     ):
         # Read the template file
         submission_content = await script.read()
@@ -148,7 +142,6 @@ class LeaderboardSubmitCog(app_commands.Group):
                 return
 
             github_command = github_cog.run_github
-            print(github_command)
             try:
                 github_thread = await github_command.callback(
                     github_cog,
@@ -180,7 +173,11 @@ class LeaderboardSubmitCog(app_commands.Group):
                     "submission_score": score,
                 })
 
-            user_id = interaction.user.global_name if interaction.user.nick is None else interaction.user.nick
+            user_id = (
+                interaction.user.global_name
+                if interaction.user.nick is None
+                else interaction.user.nick
+            )
             await interaction.followup.send(
                 "Successfully ran on GitHub runners!\n"
                 + f"Leaderboard '{leaderboard_name}'.\n"
@@ -195,10 +192,35 @@ class LeaderboardSubmitCog(app_commands.Group):
             )
 
 
+class GPUSelectionView(ui.View):
+    def __init__(self, available_gpus: list[str]):
+        super().__init__()
+
+        # Add the Select Menu with the list of GPU options
+        select = ui.Select(
+            placeholder="Select GPUs for this leaderboard...",
+            options=[SelectOption(label=gpu, value=gpu) for gpu in available_gpus],
+            min_values=1,  # Minimum number of selections
+            max_values=len(available_gpus),  # Maximum number of selections
+        )
+        select.callback = self.select_callback
+        self.add_item(select)
+
+    async def select_callback(self, interaction: Interaction):
+        # Retrieve the selected options
+        select = interaction.data["values"]
+        self.selected_gpus = select
+        await interaction.response.send_message(
+            f"Selected GPUs: {', '.join(self.selected_gpus)}",
+            ephemeral=True,
+        )
+        self.stop()
+
+
 class LeaderboardCog(commands.Cog):
     def __init__(self, bot):
         self.bot: commands.Bot = bot
-        self.get_leaderboards = bot.leaderboard_group.command(name="get")(
+        self.get_leaderboards = bot.leaderboard_group.command(name="list")(
             self.get_leaderboards
         )
         self.leaderboard_create = bot.leaderboard_group.command(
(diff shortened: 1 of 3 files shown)


src/discord-cluster-manager/utils.py
```diff
@@ -58,6 +59,18 @@ async def get_user_from_id(id, interaction, bot):
...
+def discord_followup_wrapper(interaction: discord.Interaction, msg: str) -> None:
+    """
+    To get around response messages in slash commands that are
+    called externally, send a message using the followup.
+    """
+    try:
+        await interaction.response.send_message(msg)
+
+    except Exception:
+        await interaction.followup.send(msg)
```
b9r5, line 71: I wonder if this would be better written as the following?
```suggestion
    if not interaction.response.is_done():
        await interaction.followup.send(msg)
    else:
        await interaction.response.send_message(msg)
```
We do call `interaction.response.defer()` when we intend to use followup. And that call to `defer()` ends up making `is_done()` return True.

What makes me uncomfortable about `except Exception` is that we don't really know what's gone wrong.

Honestly, I think it would work either way, but I think I prefer my suggestion if it works.

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -114,8 +114,10 @@ def create_leaderboard(self, leaderboard: LeaderboardItem):
             )
             self.connection.commit()
         except psycopg2.Error as e:
-            print(f"Error during leaderboard creation: {e}")
             self.connection.rollback()  # Ensure rollback if error occurs
+            return f"Error during leaderboard creation: {e}"
```
b9r5, line 118: I'm nervous about returning the value with the stringified exception in it. The reason is that I'm not sure what's going to be in `e`. I wouldn't expect it to include sensitive information (like the database URL), but it could. What I would suggest is to log the full string (`f"Error during leaderboard creation: {e}"`) but only return `"Error during leaderboard creation"`.

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
...
                 })
 
-            await interaction.response.send_message(
+                if err:
+                    print(err)
+                    if "duplicate key" in err:
+                        await interaction.followup.send(
+                            f'Error: Tried to create a leaderboard "{leaderboard_name}" that already exists.',
+                            ephemeral=True,
+                        )
```
b9r5, line 321: This seems fine. You checked that `"duplicate key"` is in the error?

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
...
+                err = db.create_leaderboard({
                     "name": leaderboard_name,
                     "deadline": date_value,
                     "reference_code": template_content.decode("utf-8"),
+                    "gpu_types": view.selected_gpus,
                 })
 
-            await interaction.response.send_message(
+                if err:
+                    print(err)
```
b9r5, line 316:
```suggestion
```
...or log the error.

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
...
+                    if "duplicate key" in err:
+                        await interaction.followup.send(
+                            f'Error: Tried to create a leaderboard "{leaderboard_name}" that already exists.',
+                            ephemeral=True,
+                        )
+                    else:
+                        # Handle any other errors
+                        await interaction.followup.send(
```
b9r5, line 324: Seems like a good idea to log the error in this case, so that we can look at the logs and then add more informative error messages.

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
...
-            )
+
+        except Exception as e:
+            print(str(e))
+            # Handle specific leaderboard name conflict
+            if "already exists" in str(e):
+                await interaction.followup.send(
+                    str(e),
+                    ephemeral=True,
+                )
```
b9r5, line 342: How could we get here? Shouldn't this be handled in or near the `"duplicate key" area above?

(I know I'm sounding like a broken record but) we shouldn't send `str(e)` to the Discord user. We should log a message containing `e` and send a generic message to the user like `"Leaderboard already exists"`.

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
...
+                    date_value = datetime.strptime(deadline, "%Y-%m-%d")
+                except ValueError:
+                    raise ValueError(
+                        "Invalid date format. Please use YYYY-MM-DD or YYYY-MM-DD HH:MM"
+                    )
 
+            # Ask the user to select GPUs
+            view = GPUSelectionView([gpu.name for gpu in GitHubGPU])
+
+            await interaction.response.send_message(
```
b9r5, line 286: Shouldn't this be `interaction.followup.send(...)`. If there are going to be more messages in this interaction, then `interaction.response.send_message(...)` will cause problems.

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -251,76 +273,132 @@ async def leaderboard_create(
             try:
                 date_value = datetime.strptime(deadline, "%Y-%m-%d %H:%M")
             except ValueError:
-                # If that fails, try parsing just the date (will set time to 00:00)
-                date_value = datetime.strptime(deadline, "%Y-%m-%d")
+                try:
+                    date_value = datetime.strptime(deadline, "%Y-%m-%d")
+                except ValueError:
+                    raise ValueError(
```
b9r5, line 279: I think we should do `interaction.response.send_message("Invalid date format. Please use...")` and then return here. It's a simpler control flow.

src/discord-cluster-manager/utils.py
alexzhang13: I've changed this, I think this makes sense. Will verify that this works locally before I push though.

alexzhang13: I also changed the logic for use_followup=True in the followup commit.

src/discord-cluster-manager/leaderboard_db.py
alexzhang13: Good point, for now I won't log it (we can change this later) but will remove error.

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: I'm not 100% sure what you mean by this, do you mean we sanitize the "err" string even if we never print it out?

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Removed these, this should not have been in the commit. I was debugging at the time, thanks for catching!

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Yeah this is fair, I've added logging information for the specific error here, but the Discord bot produces a fixed string.

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: I think this has to do with why I ended up re-factoring create_leaderboard to return an error string, because it wasn't being captured in the try except. I've removed this, but also made sure to remove the print errors on Discord and add them to a logger instead.

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: I've wrapped this around our followup_wrapper util to avoid this issue.

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Updated this, there was also a mistake in here regarding bad implementations I forgot to delete. It's been simplified.

b9r5:
Hey Alex, thanks for the work on this PR. I've taken another look at the changes and wanted to share my thoughts.

First, generally, from my experience I've found that it's easy to unintentionally break things in this discord.py implementation. I've been trying to stick to small, incremental changes that are quicker to review, and easier to test (and rollback if needed). While it's early days and manual testing is still necessary, I've found this approach helps reduce the testing burden and ensure quicker write/deploy/test cycles.

Given that, my recommendation is to break this PR into smaller pieces:

a. worflow.yml changes. Removing the processing here is a good improvement and ties into getting `/verifyruns` working again. I was wondering if we should use the text `none` as a placeholder value for missing scores instead of removing the functionality, but I'm also OK with this approach.

b. send_discord_message. I'm not sure about the approach and think it would be helpful to get Siro's input before deciding on the best path forward.

c. Whitespace and string style differences. The changes here make me wonder if we should standardize on formatting. Has this been discussed yet? I'm open to a discussion on this to avoid inconsistency (and any poor style choices on my part 🙂) without spending too much time on formatting reviews.

d. GPU selection view changes. These look quite cool.

msaroufim:
On styling lemme just use ruff and do a one time lint of the repo

PR #59, opened 2024-12-17
Replace all relevant Interaction.send/followups with send_discord_message wrapper

alexzhang13:
## Description

To avoid issues where Discord complains about the correct usage of send / followup, we add a wrapper that checks which is valid, and uses the correct command. This feature is important for callbacks, where an existing response might exist, despite the GitHub slash command function assuming it does not.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/cogs/leaderboard_cog.py
+++ b/src/discord-cluster-manager/cogs/leaderboard_cog.py
@@ -5,7 +5,7 @@ import discord
 from consts import GitHubGPU, ModalGPU
 from discord import Interaction, SelectOption, app_commands, ui
 from discord.ext import commands
-from utils import extract_score, get_user_from_id, setup_logging
+from utils import extract_score, get_user_from_id, setup_logging, send_discord_message
 
 logger = setup_logging()
 
@@ -148,7 +148,6 @@ class LeaderboardSubmitCog(app_commands.Group):
                     script,
                     gpu_type,
                     reference_code=reference_code,
-                    use_followup=True,
                 )
             except discord.errors.NotFound as e:
                 print(f"Webhook not found: {e}")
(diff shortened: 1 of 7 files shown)


src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -335,11 +339,11 @@ async def leaderboard_create(
...
             # Handle any other errors
-            await interaction.followup.send(
+            await send_discord_message(
+                interaction,
                 "Error in leaderboard creation.",
                 ephemeral=True,
             )
-            
 
     @discord.app_commands.describe(leaderboard_name="Name of the leaderboard")
```
S1ro1, line 348: @alexzhang13 I think we should just make it use our `send_discord_message` in all cases, so would replace these `interaction.response.send_message` with it as well

src/discord-cluster-manager/cogs/leaderboard_cog.py
S1ro1, line 1: I'd replace all occuranges of the discord default way with our wrapper, sadly I cannot mark those as Github doesn't allow comments on lines outside of PR, but in methods `submit_modal` and `get_leaderboard_submission`

src/discord-cluster-manager/cogs/verify_run_cog.py
S1ro1, line 1: Also in here, would change the occurances, just so we're doing it all the same way

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Oh there's a weird case with embed=embed or whatever where it doesn't exactly work, but otherwise I can replace the rest.

S1ro1: what's the issue with `embed=embed`? Does one of the `followup.send` or `interaction.response.send_message` not support it?

alexzhang13: Yeah they expect a message, but I can change this as well.

alexzhang13: We can work around this in a later PR though, for now I think we know what to expect for the embeds (it's always a followup) so we can keep it as is.

S1ro1 (approved):
lgtm

PR #66, opened 2024-12-22
remove unused DB environment variables

b9r5:
## Description

This PR removes some old constants whose values came from DB environment variables at one point. We only use DATABASE_URL at this point, so the old constants can be removed.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


b9r5:
@alexzhang13 and I decided not to merge this because he still uses the env vars to test on his local. Postgres can be tricky to configure, and using the URL may be nontrivial for some developers. We think we should just leave the environment variables in.

PR #76, opened 2024-12-26
Pytorch Profiler

AndreSlavescu:
## Description

Please provide a brief summary of the changes in this pull request.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/cogs/github_cog.py
+++ b/src/discord-cluster-manager/cogs/github_cog.py
@@ -241,3 +241,86 @@ class GitHubCog(commands.Cog):
             return "No training artifacts found"
         except Exception as e:
             return f"Error downloading artifacts: {str(e)}"
+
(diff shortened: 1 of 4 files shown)


alexzhang13:
@AndreSlavescu The only problem I see with this is that it implicitly just wraps a profiler around the top of a file, but this can be problematic if someone just submits a standard kernel file that's formatted as a bunch of functions (you'll have to be explicit about this). Like basically this only works if someone submits a file that doesn't run.

I need to give this more thought, @S1ro1 maybe you have better thoughts.

Can you instead make sure for now that profiling function goes into say the if __name__ == "__main__"?

alexzhang13 (changes requested):
For now, can you make it so the PyTorch profiler goes inside say `if __name__ == '__main__'`? also make sure it indents properly so it actually runs.

PR #81, opened 2024-12-26
Feat/uv

S1ro1:
## Description

Moves the whole project to use `uv` instead of vanilla pip, it makes development process easier with a price of small changes to heroku deploy (adding a preprocessing buildpack as heroku doesn't natively support uv yet)

Also moves the whole configuration to `pyproject.toml`, enabling development dependencies and also configuring ruff in one place.

All updates are reflected in the readme.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


uv.lock
```diff
@@ -0,0 +1,1024 @@
+version = 1
```
msaroufim, line 1: do we need to check this file in?

runtime.txt
```diff
@@ -1 +0,0 @@
-python-3.13.0
```
msaroufim, line 1: IIRC this was needed for heroku not for the package

msaroufim (changes requested):

msaroufim:
SHould also update the procfile cause that's what heroku will run

uv.lock
S1ro1: I'm not sure on this one to be fair, as heroku doesn't support uv by default, we need a preprocessor to generate requirements.txt from uv.lock [here](https://github.com/dropseed/heroku-buildpack-uv?tab=readme-ov-file)

I remember from when I used poetry, having .lock in VCS was the correct practice when producing an executable, not a library.

runtime.txt
S1ro1: As mentioned in the previous reply, this gets generated via the buildpack on heroku side.

S1ro1:
https://github.com/heroku/heroku-buildpack-python/issues/1616 states that uv has no full support for uv, so basically what this does is just uses uv for the development, heroku would still use the default python worker. We can also stick with the current way in favor of having easier deploy and not having uv for personal development @msaroufim .

msaroufim:
Yeah unfortunately without heroku having better support for uv I feel like this PR adds more complexity for some benefit and I don't feel like it's super worth it yet

PR #82, opened 2024-12-26
Add support for submitting to specific GPUs per leaderboard and show

alexzhang13:
## Description
This PR adds functionality for submitting to a leaderboard with certain GPUs, and showing the results (these go hand-in-hand currently, so I didn't separate the PRs).

As an example:
Screenshot 2024-12-26 at 6 26 33 PM

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/leaderboard_db.py
+++ b/src/discord-cluster-manager/leaderboard_db.py
@@ -1,4 +1,4 @@
-from typing import Optional, List
+from typing import List, Optional
 
 import discord
 import psycopg2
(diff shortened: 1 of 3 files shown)


S1ro1:
I personally would go with showing 1 table, but separating them via a column or something, this would properly showcase if some gpus are way faster than other etc. Up to discuss this also with @msaroufim

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -198,17 +219,20 @@ def get_leaderboard(self, leaderboard_name: str) -> int | None:
             return None
 
     # TODO: add GPU type
-    def get_leaderboard_submissions(self, leaderboard_name: str) -> list[SubmissionItem]:
+    def get_leaderboard_submissions(
+        self, leaderboard_name: str, gpu_name: str
```
S1ro1, line 223: Can we maybe make this optional? Or create a new method that filters based on key: value pairs, etc. I would like this as pure "get all"

S1ro1 (changes requested):

alexzhang13:
> I personally would go with showing 1 table, but separating them via a column or something, this would properly showcase if some gpus are way faster than other etc. Up to discuss this also with @msaroufim

Eh I still think comparing across GPUs for a kernel is an apples to oranges thing. I like them being completely separate leaderboards, and being displayed this way as well.

src/discord-cluster-manager/leaderboard_db.py
alexzhang13: But get all is weird -- if you want to get all you can just select all GPUs, otw I don't like this being a default option. Because the selection is in the UI now, I like the the user having to select what GPUs they want.

S1ro1 (approved):
lgtm

PR #85, opened 2024-12-28
private gpu view

AndreSlavescu:
## Description

Please provide a brief summary of the changes in this pull request.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


alexzhang13 (commented):
I don't know the actual behavior of this, @S1ro1 probably will understand better. But my understanding is that either:
1) The interaction happens relative to the person interacting. So in this case these if statements don't do anything.
2) The interactions happen per user, in which case this seems fine? But then I am afraid these if conditions trigger for every user, which is also bad. 

If we can confirm this works as intended without performance issues I'm happy to use this for now.

PR #89, opened 2024-12-31
Update README with new reference script expectations for Python / CUDA.

alexzhang13:
## Description

Add reference kernel function signatures for CUDA, also update Python reference kernel function signatures.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,24 +253,25 @@ analysis for the leaderboard. The difficult of CUDA evaluation scripts is we nee
 handle the typing system for tensors. The `reference.cu` that the leaderboard creator submits must have
 the following function signatures with their implementations filled out:
 
-The main difference is we now need to define an alias for the type that the input / output list of
-tensors take. A simple solution is to pre-define an array of `const int` sizes, then define an array
-of containers, e.g. `std::array<std::vector<float>, N_SIZES>`.
+The main difference is we now need to define an alias for the type that the input / outputs are. A
+simple and common example is a list of FP32 tensors, which can be defined using a pre-defined array of 
+`const int`s called `N_SIZES`, then define an array of containers, e.g. 
+`std::array<std::vector<float>, N_SIZES>`.
 
 ```cuda
-// (List of) Tensor types as input, e.g. using input_t = std::array<std::vector<float>, IN_SIZES>;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -240,11 +240,47 @@ def ref_kernel(input: torch.Tensor) -> torch.Tensor:
...
+
+The main difference is we now need to define an alias for the type that the input / output list of
+tensors take. A simple solution is to pre-define an array of `const int` sizes, then define an array
+of containers, e.g. `std::array<std::vector<float>, N_SIZES>`.
+
+```cuda
+// (List of) Tensor types as input, e.g. using input_t = std::array<std::vector<float>, IN_SIZES>;
+using input_t = ...;
+
+// (List of) Tensor types as outputs, e.g. using output_t = std::array<std::vector<float>, OUT_SIZES>;
```
S1ro1, line 264: This would be effectively the same, I would name IN/OUT_SIZES differently to denote that it's the number of subproblems/inputs. Maybe N_INPUTS? or something as such.

README.md
```diff
@@ -240,11 +240,47 @@ def ref_kernel(input: torch.Tensor) -> torch.Tensor:
...
+using output_t = ...;
+
+// Generate random (list of) tensors
```
S1ro1, line 267: // Generate data in format as defined above? we don't want to enforce list of tensors etc... since the creator is responsible for calling the kernel, they just get input_t

README.md
```diff
@@ -240,11 +240,47 @@ def ref_kernel(input: torch.Tensor) -> torch.Tensor:
...
+input_t generate_input() {
+    // Implement me...
+}
+
+
+// Reference kernel (over list of data), host code.
```
S1ro1, line 273: Again same case here, Over data as defined by input_t

S1ro1 (changes requested):

README.md
alexzhang13: This is true, I just left a comment above it about an example being a List of Tensors and removed these hints.

README.md
alexzhang13: This is true, I just left a comment above it about an example being a List of Tensors and removed these hints.

S1ro1 (approved):
lgtm

PR #91, opened 2025-01-03
Update Python Reference code / docs to match CUDA

alexzhang13:
## Description

@S1ro1 Update reference code and README to be generic.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,19 +230,21 @@ specify the available GPUs that the leaderboard evaluates on.
 
 The Discord bot internally contains an `eval.py` script that handles the correctness and timing
 analysis for the leaderboard. The `reference_code` that the leaderboard creator submits must have
-the following function signatures with their implementations filled out:
+the following function signatures with their implementations filled out. `InputType` and
+`OutputType` are generics that could be a `torch.Tensor`, `List[torch.Tensor]`, etc. 
+depending on the reference code specifications. We leave this flexibility to the leaderboard creator.
 
 ```python
 # Reference kernel implementation.
-def ref_kernel(input: torch.Tensor) -> torch.Tensor:
+def ref_kernel(input: InputType) -> OutputType:
     # Implement me...
 
 # Generate a list of tensors as input to the kernel
-def generate_input() -> List[torch.Tensor]:
+def generate_input() -> InputType:
     # Implement me...
 
 # Verify correctness of reference and output
-def check_implementation(input: torch.Tensor, output: torch.Tensor) -> bool:
+def check_implementation(custom_out: OutputType, reference_out: OutputType) -> bool:
     # Implement me...
 ```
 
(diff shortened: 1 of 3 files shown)


src/discord-cluster-manager/leaderboard_eval.py
```diff
@@ -28,30 +26,28 @@ def metric():
...
-    print('timing custom implementation...')
-    input_tensor = generate_input()
+    # Timing Code
+    input_tensors = generate_input()
     start_time = time.time()
     for _ in range(timed_runs):
-        for input in input_tensors:
-            _ = custom_kernel(input)
+        _ = custom_kernel(input_tensors)
+    end_time = time.time()
```
msaroufim, line 41: This is missing a cuda sync, i.e the timings will only be correct if the kernel code ends by doing the cuda sync itself by for example sending tensors back to cpu to print them

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -168,6 +168,10 @@ async def submit_github(
         leaderboard_name: str,
         script: discord.Attachment,
     ):
+        # Don't allow submissions if deadline is past
```
msaroufim, line 171: do we need deadlines even explicitly as part of the flow? feels fine to get new submissions as long as they're all timestamped

README.md
```diff
@@ -289,11 +291,13 @@ bool check_implementation(output_t out, output_t ref) {
...
 
-The leaderboard submission for _Python code_ requires the following function signatures:
+The leaderboard submission for _Python code_ requires the following function signatures, where
+`InputType` and `OutputType` are generics that could be a `torch.Tensor`, `List[torch.Tensor]`, 
+etc. depending on the reference code specifications.
 
 ```python
 # User kernel implementation.
-def custom_kernel(input: torch.Tensor) -> torch.Tensor:
+def custom_kernel(input: InputType) -> OutputType:
```
msaroufim, line 300: why not a union type of torch.tensor and list[torch.tensor]?

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: We don't need them, but the rest of the workflow has them atm which is why this is here. I'm fine with getting rid of this TBH, but we'd do this in a different PR.

README.md
alexzhang13: The idea is that this could in theory be anything. The most likely scenarios are torch.Tensor and list[torch.Tensor], but the ref code creator could pass in anything they want here (a tuple of tensor, int, float, etc.) and the submitter must obey this type. This is in-line with how we designed it for the CUDA implementation as an arbitrary struct / aliased type.

S1ro1 (approved):
lgtm

PR #92, opened 2025-01-04
Feat/UI fixes and Improvements

S1ro1:
## Description

A Huge PR that does basically everything to cleanup the leaderboard submission flow.

- Provides 2 new commands to provide help to users creating submissions
       * `/leaderboard reference-code {leaderboard_name}`
       * `/leaderboard eval-code {language}`
- Provides a way to get personal submissions to a leaderboard via `/leaderboard show-personal {leaderboard_name}

- Moves UI Elements to a separate `ui` folder to cleanup some of the files
- Provides a simple way to create tables with pagination via `ui/table` and method `create_table`
- Moves commands currently creating tables to use this table
- Remove some messages, reducing visual clutter (i.e. "Created thread for..."), and left only the default Discord message to announce this
- Move GPUSelectionView to be ephemeral by passing `ephemeral=True` to `interaction.defer`
- Make required commands ephemeral

DEMO:

image
image
image
image
image

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13, line 1: from ui.delete_confirmation import create_delete_confirmation_modal

doesn't refer to anything that exists atm

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -218,23 +218,34 @@ def get_leaderboard(self, leaderboard_name: str) -> int | None:
...
-            s.gpu_type
+        query = """
+            SELECT
+                s.name,
+                s.user_id,
+                s.code,
+                s.submission_time,
+                s.score,
+                s.gpu_type,
+                RANK() OVER (PARTITION BY s.gpu_type ORDER BY s.score ASC) as rank
```
msaroufim, line 232: worth a comment - also i forget how y'all defined scores, we are mostly ranking on time. So is score = 1/ time?

src/discord-cluster-manager/utils.py
```diff
@@ -82,16 +83,16 @@ def extract_score(score_str: str) -> float:
         return None
 
 
-async def display_lb_submissions(interaction, bot, leaderboard_name: str, gpu: str):
+async def display_lb_submissions(
```
msaroufim, line 86: nit: should this function be in leaderboard_cog? there's a lot more files i need to click around now

src/discord-cluster-manager/ui/gpu_selection.py
```diff
@@ -0,0 +1,28 @@
+from discord import Interaction, SelectOption, ui
```
msaroufim, line 1: does this need its own standalone file?

src/discord-cluster-manager/ui/gpu_selection.py
```diff
@@ -0,0 +1,28 @@
...
+    async def select_callback(self, interaction: Interaction):
+        # Retrieve the selected options
+        select = interaction.data["values"]
+        self.selected_gpus = select
+        # Acknowledge the interaction
+        await interaction.response.defer(ephemeral=True)
+        self.stop()
+
+
+def create_gpu_selection_view(available_gpus: list[str]) -> GPUSelectionView:
```
msaroufim, line 27: not a fan of shallow functions like this, they justt add indirection and make things harder to understand

src/discord-cluster-manager/ui/table.py
```diff
@@ -0,0 +1,93 @@
...
+    table_rows.append(separator)
+
+    for item in page_data:
+        row = "  ".join(str(item[header]).ljust(col_widths[header]) for header in headers)
+        table_rows.append(row)
+
+    return discord.Embed(description=f"```\n{'\n'.join(table_rows)}\n```")
+
+
+def create_table(
```
msaroufim, line 71: another shallow function, although could be convinced this is helpful to keep

src/discord-cluster-manager/ui/delete_confirmation.py
```diff
@@ -0,0 +1,46 @@
+import discord
```
msaroufim, line 1: Does this do deletions for both individual entries and leaderboards?

Also does this mean anyone can delete other messages? If so we need a notion of credentialed user

I would prefer if deleting specific leaderboards was really about deleting a file on github

And as far as deleting entries, i feel like we can take those requests ad hoc, so basically someone gives us a run ID and we can delete it from the DB cc @b9r5

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -526,8 +517,28 @@ async def get_leaderboard_submissions(
                     interaction, "An unknown error occurred.", ephemeral=True
                 )
 
+    @discord.app_commands.describe(leaderboard_name="Name of the leaderboard")
+    @app_commands.autocomplete(leaderboard_name=leaderboard_name_autocomplete)
```
msaroufim, line 521: autocomplete fancy!

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -355,35 +306,68 @@ async def get_leaderboards(self, interaction: discord.Interaction):
...
+
+        with self.bot.leaderboard_db as db:
+            leaderboard_item = db.get_leaderboard(leaderboard_name)
+            if not leaderboard_item:
+                await send_discord_message(interaction, "Leaderboard not found.", ephemeral=True)
+                return
+
+        code = leaderboard_item["reference_code"]
+        code_file = StringIO(code)
+        language = "cpp" if "#include" in code else "py"
```
msaroufim, line 337: seems brittle :P

msaroufim (changes requested):

src/discord-cluster-manager/utils.py
S1ro1: I would also move this to leaderboard_cog honestly and make it a method, was there before so I kept it.

src/discord-cluster-manager/ui/delete_confirmation.py
S1ro1: For now, this was done to delete the entries as well, as else it would keep db in a wrongful state since we have it connected via foreign keys. I would say this is more like an option if someone creates something wrong. 

Yes, credentials is my next to do.

src/discord-cluster-manager/ui/gpu_selection.py
S1ro1: For all of the `ui` files I wanted to make it a standalone lib with __init__.py, but it creates a circular dependency I am not able to solve, so I'd be ok with just creating `ui.utils` where we have misc ui utils as such.

src/discord-cluster-manager/cogs/leaderboard_cog.py
S1ro1: Yeah, currently best way we can do it (unless smarter heuristic), but we should probably add type to leaderboard table.

src/discord-cluster-manager/ui/table.py
S1ro1: Can't actually make this a class-only, as we do return a standalone embed in case of no data, which I like since we don't need random checks all around the place.

src/discord-cluster-manager/leaderboard_db.py
S1ro1: Currently it's not even 1/time lol, it's just straight up time, but it's a small fix that we have forgot about.

src/discord-cluster-manager/ui/table.py
alexzhang13: ^^ I think this is actually fine, worst case scenario we can just move this somewhere else in some utils file where it's all shallow functions

alexzhang13 (approved):

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Yeah there are defo weird cases to consider (say if Python code had some weird string like leaderboard_eval.py where you include cpp code) but I think we can separate this PR, and just add type to LB table.

alexzhang13: We should just keep this in mind for later.

alexzhang13:
Screenshot 2025-01-04 at 11 01 20 PM

This code is now updated as well to return long code files as a file instead of Discord text to avoid complaints about length.

msaroufim (approved):

PR #93, opened 2025-01-04
Requirement set up and 

simonguozirui:
## Description

Please provide a summary of the changes in this pull request.
* Disable `audioop-lts` pip package as not sure why it is needed. We like to use Python `3.11` but this makes it use `3.13`
* Get rid of SSL requirements for local dev server set up. Mayb we need this for staging, but be good to not require this in debug mode?

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 PyGithub
 aiohttp
 discord.py
-audioop-lts # discord.py imports using * syntax 
+# audioop-lts # discord.py imports using * syntax 
 python-dotenv
 requests
 modal
--- a/src/discord-cluster-manager/leaderboard_db.py
+++ b/src/discord-cluster-manager/leaderboard_db.py
@@ -52,7 +52,7 @@ class LeaderboardDB:
         """Establish connection to the database"""
         try:
             self.connection = (
-                psycopg2.connect(DATABASE_URL, sslmode="require")
+                psycopg2.connect(DATABASE_URL)
                 if DATABASE_URL
                 else psycopg2.connect(**self.connection_params)
             )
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 PyGithub
 aiohttp
 discord.py
-audioop-lts # discord.py imports using * syntax 
+# audioop-lts # discord.py imports using * syntax
```
msaroufim, line 4: this would break our heroku deployment, so you'd need to also update https://github.com/gpu-mode/discord-cluster-manager/blob/main/runtime.txt

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -52,7 +52,7 @@ def connect(self) -> bool:
         """Establish connection to the database"""
         try:
             self.connection = (
-                psycopg2.connect(DATABASE_URL, sslmode="require")
+                psycopg2.connect(DATABASE_URL)
                 if DATABASE_URL
```
msaroufim, line 56: why tho? This would disable ssl everywhere

requirements.txt
simonguozirui: I see, let's stick with python 3.11 then.

src/discord-cluster-manager/leaderboard_db.py
simonguozirui: We disabled this while @alexzhang13  and I were trying to dev locally otherwise it throw errors. Maybe a better way is to have a local debug mode that does not require SSL.

simonguozirui:
Closing this for now as we understand now why these code choices were made.

PR #97, opened 2025-01-06
Add Modal support for Leaderboard (Python + CUDA)

alexzhang13:
## Description

Adds Python and CUDA support for Modal runners on the leaderboard. This will be especially useful for reducing overhead that the GitHub runners have, and provides a wider variety of devices to use. Also, this PR is in preparation for CUTLASS and TK support for the leaderboard.
 
Screenshot 2025-01-06 at 12 19 29 AM

Also includes the Modal GPUs for `/leaderboard create ...`
image

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


src/discord-cluster-manager/env.py
alexzhang13, line 1: You may be confused why this file is separated from `consts.py`. The main reason is that I wanted to have paths defined as constants for where certain LB files sit in Modal, but when you try to use them, the Modal runner now expects to import from this file. If you import the old `consts.py` file, it expects the Modal environment to have all the right ENV variables that our discord bot has, which makes no sense. The consts.py file I felt should just be stand-alone consts, while the `env.py` file is environment variables and any relevant initialization functions.

src/discord-cluster-manager/modal_runner.py
```diff
@@ -96,46 +179,75 @@ def run_cuda_script(script_content: str, timeout_seconds: int = 600) -> tuple[st
...
+                # these check cuda compiler is also available
+                subprocess.run(["nvcc", "--version"], check=True)
+                subprocess.run(["which", "nvcc"], check=True)
+            except Exception:
+                return "nvcc not found.", 0.0
+
+            NVCC_FILES = "eval.cu"
+            # Write submission files to directory
+            if reference_content is not None:
+                with open("/root/reference.cuh", "w") as f:
```
S1ro1, line 198: Let's make this consistent between python and cuda, I think we can create all of the files in the local directory (not in /root) and not mount anything as we create them only inside the container. That would make this whole process isolated from other submissions as well.

src/discord-cluster-manager/modal_runner.py
```diff
@@ -96,46 +179,75 @@ def run_cuda_script(script_content: str, timeout_seconds: int = 600) -> tuple[st
...
         return f"Timeout Error: {str(e)}", 0.0
     except Exception as e:
         return f"Error: {str(e)}", 0.0
     finally:
-        if os.path.exists("script.cu"):
-            os.remove("script.cu")
-        if os.path.exists("script.out"):
-            os.remove("script.out")
-        sys.stdout = sys.__stdout__
+        if os.path.exists("reference.cuh"):
```
S1ro1, line 246: nit: possible just iterate over list of strings: for f in files: if f exists -> os.remove(f), this looks ugly and introduces potential issues in typos etc, same for the run_pytorch

S1ro1 (changes requested):

src/discord-cluster-manager/modal_runner.py
alexzhang13: Good point, fixed this

src/discord-cluster-manager/modal_runner.py
alexzhang13: Fixed this as well.

S1ro1 (approved):
lgtm

PR #99, opened 2025-01-07
Finish up Modal support + Example of ThunderKittens working

alexzhang13:
## Description

This is a relatively large PR with a lot of finishing changes for modal support and ThunderKittens (TK) support.

Namely,

1. We can compile (only tested on just importing and building) ThunderKittens kernel on Modal. More extensive testing coming.
2. The UI for Modal actually works now, and the GPUs you submit to are correct. You can now submit to whatever GPUs you want. You can also submit to more than just the T4s now.
3. A big re-factor to greatly simplify Modal runner code, and merge most of the GitHub runner logic to also work with Modal. The GH/Modal runner logic is now <30 lines.
4. We currently add clone TK when setting up the Modal image, but we may want to do this differently. Happy to make changes in the future.
5. The Modal runners run the Modal Cog, similar to the GitHub runner. This helps a lot with modularity and removing the need to design two different functions for running code on Modal.

Some potential issues for the future:
* I frequently run into "Modal GPUs not hydrated" when starting up the bot. Not sure if this just means the resources aren't ready, but this issue goes away quickly.
* The way we implement multiple modal GPU runners is really disgusting, and it has to do with the function wrappers. I actually had multiple attempts at this that didn't work, but it's hard to distinguish between the issue above and an actual issue. I will return to this later, as I think the current solution is unsatisfactory (we should still just push it for now because it "works").

## Screenshots
Screenshot 2025-01-07 at 12 45 24 AM

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/cogs/leaderboard_cog.py
+++ b/src/discord-cluster-manager/cogs/leaderboard_cog.py
@@ -168,6 +168,10 @@ class LeaderboardSubmitCog(app_commands.Group):
         leaderboard_name: str,
         script: discord.Attachment,
     ):
+        # Don't allow submissions if deadline is past
+        with self.bot.leaderboard_db as db:
+            leaderboard_item = db.get_leaderboard(leaderboard_name)
+
         # Read the template file
         submission_content = await script.read()
 
@@ -183,8 +187,19 @@ class LeaderboardSubmitCog(app_commands.Group):
             # Read and convert reference code
             reference_code = None
             with self.bot.leaderboard_db as db:
-                # TODO: query that gets reference code given leaderboard name
                 leaderboard_item = db.get_leaderboard(leaderboard_name)
+
+                now = datetime.now()
+                deadline = leaderboard_item["deadline"]
+
+                if now.date() > deadline.date():
+                    await send_discord_message(
+                        interaction,
+                        f"The deadline to submit to {leaderboard_name} has passed.\n"
+                        + f"It was {deadline.date()} and today is {now.date()}.",
+                    )
+                    return
+
                 if not leaderboard_item:
                     await send_discord_message(
                         interaction,
--- a/src/discord-cluster-manager/cogs/leaderboard_cog.py
+++ b/src/discord-cluster-manager/cogs/leaderboard_cog.py
@@ -60,36 +60,50 @@ class LeaderboardSubmitCog(app_commands.Group):
 
         message_contents = [msg.content async for msg in discord_thread.history(limit=None)]
 
-        # Compute eval or submission score, call runner here.
-        # TODO: Make this more robust later
-        score = extract_score("".join(message_contents))
+        try:
+            # For CUDA leaderboards, make more robust
+            if "check_implementation failed" in message_contents:
+                raise Exception("check_implementation failed. User kernel is incorrect.")
 
-        with self.bot.leaderboard_db as db:
-            db.create_submission(
-                {
-                    "submission_name": script.filename,
-                    "submission_time": datetime.now(),
-                    "leaderboard_name": leaderboard_name,
-                    "code": submission_content,
-                    "user_id": interaction.user.id,
-                    "submission_score": score,
-                    "gpu_type": gpu,
-                }
+            # Compute eval or submission score, call runner here.
+            # TODO: Make this more robust later
+            score = extract_score("".join(message_contents))
+
+            with self.bot.leaderboard_db as db:
+                db.create_submission(
+                    {
+                        "submission_name": script.filename,
+                        "submission_time": datetime.now(),
+                        "leaderboard_name": leaderboard_name,
+                        "code": submission_content,
+                        "user_id": interaction.user.id,
+                        "submission_score": score,
+                        "gpu_type": gpu,
+                    }
+                )
+
+            user_id = (
+                interaction.user.global_name
+                if interaction.user.nick is None
+                else interaction.user.nick
             )
 
-        user_id = (
-            interaction.user.global_name if interaction.user.nick is None else interaction.user.nick
-        )
-
-        await send_discord_message(
-            interaction,
-            f"Successfully ran on {gpu} using {runner_name} runners!\n"
-            + f"Leaderboard '{leaderboard_name}'.\n"
-            + f"Submission title: {script.filename}.\n"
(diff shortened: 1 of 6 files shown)


src/discord-cluster-manager/modal_runner.py
```diff
@@ -223,22 +232,30 @@ def run_cuda_script(  # # noqa: C901
...
+                    score,
+                )  # To make sure error is thrown on LB
 
             return run_process.stdout, score
 
-    except TimeoutException as e:
-        return f"Timeout Error: {str(e)}", 0.0
-    except Exception as e:
-        return f"Error: {str(e)}", 0.0
+    except TimeoutException:
```
msaroufim, line 255: how come? in the case of an exception we lose the error completely now

src/discord-cluster-manager/modal_runner_archs.py
```diff
@@ -0,0 +1,199 @@
...
+    return run_pytorch_script(
+        script_content,
+        reference_content,
+        submission_content,
+        timeout_seconds,
+        arch=80,  # Targeting L4 (sm_80)
+    )
+
+
+# A100: sm_80 (Ampere architecture)
```
msaroufim, line 85: make a dict for enum for SM to GPU mapping otherwise people will need to do a google search first

src/discord-cluster-manager/modal_runner_archs.py
```diff
@@ -0,0 +1,199 @@
...
+    "t4": run_cuda_script_t4,
+    "l4": run_cuda_script_l4,
+    "a100": run_cuda_script_a100,
+    "h100": run_cuda_script_h100,
+}
+
+
+def get_pytorch_modal_runner(gpu_type: str):
+    """
+    Returns the appropriate PyTorch function for the given gpu_type.
```
msaroufim, line 180: comments for these functions are not informative

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -24,159 +25,49 @@
...
+    async def async_submit_cog_job(
         self,
         interaction: discord.Interaction,
         leaderboard_name: str,
         script: discord.Attachment,
+        command,
+        reference_code,
+        submission_content,
+        cog: commands.Cog,
+        gpu: str,
```
msaroufim, line 42: string typed inputs is not infromative, you can make this an enum as well

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -24,159 +25,49 @@
...
-                        eval_code,
-                        reference_content=reference_code,
-                        submission_content=submission_content.decode("utf-8"),
-                    )
-                else:
-                    from modal_runner import run_pytorch_script
+        try:
+            # For CUDA leaderboards, make more robust
+            if "check_implementation failed" in message_contents:
+                raise Exception("check_implementation failed. User kernel is incorrect.")
```
msaroufim, line 66: naked exceptions are bad, these will catch anything! If you do raise a naked exception, log an informative error cause here we won't know how to debug

For example in this case if the outputs don't match then raise that specific error and show both outputs

src/discord-cluster-manager/cogs/leaderboard_cog.py
```diff
@@ -24,159 +25,49 @@
...
-                        eval_code,
-                        reference_content=reference_code,
-                        submission_content=submission_content.decode("utf-8"),
-                    )
-
-            if not all([modal_cog]):
-                await send_discord_message(interaction, "❌ Required cogs not found!")
-                return
+            # Compute eval or submission score, call runner here.
+            # TODO: Make this more robust later
```
msaroufim, line 69: +1

src/discord-cluster-manager/cogs/leaderboard_cog.py
simonguozirui: good idea

src/discord-cluster-manager/modal_runner.py
```diff
@@ -23,30 +38,28 @@
...
+        "numpy",
+    )
+    .run_commands(
+        "update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100 "
+        + "--slave /usr/bin/g++ g++ /usr/bin/g++-11",
+        # "apt update",
+        # "apt  -y install clang-10", # this should be clang-10 but I can't get it to work yet
+        #
+        "git clone https://github.com/HazyResearch/ThunderKittens.git",
+        # "cd /ThunderKittens && pwd && python setup.py install",
```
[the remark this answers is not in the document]
simonguozirui: Right now no, you just need to link and import. You can then choose to compile TK program as binary or add python binding to call them.  

@alexzhang13 and i thought about trying storing the repo on a shared modal file system.

src/discord-cluster-manager/cogs/leaderboard_cog.py
alexzhang13: Yeah this was just bad programming, this shouldn't even be an Exception. I've changed this to return a Discord message with the appropriate information.

src/discord-cluster-manager/modal_runner_archs.py
```diff
@@ -0,0 +1,199 @@
...
+
+cuda_function_map = {
+    "t4": run_cuda_script_t4,
+    "l4": run_cuda_script_l4,
+    "a100": run_cuda_script_a100,
+    "h100": run_cuda_script_h100,
+}
+
+
+def get_pytorch_modal_runner(gpu_type: str):
```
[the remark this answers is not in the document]
alexzhang13: 1 and 3. Yeah so this was breaking before. I had implemented this exact approach but it was causing issues before, so I wanted to separate this into another PR and let this version be minimally working.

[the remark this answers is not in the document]
alexzhang13: NVM, this is all fixed now.

S1ro1:
Have a minimal viable product, further refactor of the modal runner calling logic is done in #100. For now, merging as tested locally and other comments were addressed.

simonguozirui:
Fantastic work @alexzhang13! now we can easily test on modal.

PR #110, opened 2025-01-10
Consistent naming between CUDA and Python submissions

alexzhang13:
## Description

CUDA submissions now expect the same function names (basically signatures) as Python submissions, so

reference --> ref_kernel
submission --> custom_kernel

Also, examples are changed to reflect this change (I mainly did this because we don't have any interesting active CUDA leaderboards right now, so other users on the server won't get confused). The other difference now we accept both `.cuh` and `.cpp` files, they are treated the same as `.cu` but I thought it was weird that we weren't allowing them as submissions.

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


S1ro1 (commented):
lgtm

PR #112, opened 2025-01-10
Vectors Add - Pytorch CUDA inline and Triton example

rkinas:
## Description

Implemented two sample of vectors add:
- Pytorch CUDA online 
- Triton

## Checklist

- [ X] Run the slash command `/verifyruns` on your own server.
    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,12 @@ support multiple GPU types yet)
 /leaderboard show {name: str}
 ```
 
+Display all personal scores (runtime) from a specific leaderboard.
+
+```
+/leaderboard show-personal {name: str}
+```
+
 #### GPU Kernel-specific Commands
 
 We plan to add support for the PyTorch profiler and CUDA NSight Compute CLI to allow users to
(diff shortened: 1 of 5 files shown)


alexzhang13 (approved):
These all work locally :) Let me try to fix the linter issues though. I have to actually change the GH runners to support inlining, but this I will do separately. Thanks for the awesome work!

simonguozirui:
I like the inline examples. If current infra supports problem definition and eval submission being inline torch code, then we can technically allow it to run all of the Kernelbench problems.

Kernelbench is a benchmark that I worked on with @anneouyang; Currently that has 250+ Torch task definitions that ask the model to generate Inline CUDA with Torch. 

If it is of interest, we could add support for KernelBench to be imported as leaderboard task definition easily or as leaderboards for the competition. 
Just want to throw that out there; curious to get your thoughts @alexzhang13  @msaroufim on this.

msaroufim:
@simonguozirui from the very first moment you and @anneouyang first shared your work on KernelBench I felt like the project would be most impactful if both humans and AI could compete on the tasks that you define. So let's do it :)

PR #119, opened 2025-01-11
Fix: enable creation of a lb with the same name after deleting the old one

S1ro1:
## Description

Enables creating a leaderboard with the same name after deleting the old one. Issue was caused by the forum channel already existing. Therefore we solve this by renaming the original thread to {original_name} - archived at {timestamp} to avoid having 2 same names.

image
image

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


alexzhang13 (approved):
Ah that's really funny -- I didn't expect this to be the issue! Clever solution though.

PR #120, opened 2025-01-11
factored out run_cuda_eval, so it can be used independently of modal

ngc92:
split out `run_cuda_script` to make it independent of model;
in particular, this allows you to run simple checks locally (if you have a working cuda installation), as the `scripts/local-test.py` script illustrates (cwd is expected to be the repository root).


msaroufim (approved):

msaroufim:
I still a lot of changes happening so just lmk whne you need a review

ngc92:
> I still a lot of changes happening so just lmk when you need a review

I mainly want to add a simple github CI action that runs the identity kernel to verify this is correct. Except the CI job claims that the result is wrong  :(

msaroufim (approved):

PR #126, opened 2025-01-11
Using pipe to transfer data from runner to bot

ngc92:
This PR changes the mechanism by which the eval.cu script reports back to its caller. Instead of writing to stdout, we use a dedicated pipe as our communication channel. 
This is cleaner, IMO, as it leaves the standard io channels for the user (e.g., for debugging), and somewhat safer (simple `std::cout` can no longer confuse the bot), and we can transfer more data without worrying that we accidentally pick up a user message.
As an example, in this PR we already calculate and transfer additional benchmark statistics, though they are not processed further in  the bot for now.
I'm not completely sure about  the exact design of the ` PopcornOutput` class, for now, this very simple implementation seems sufficient, though.

### DIFF
--- a/scripts/ci_test_cuda.py
+++ b/scripts/ci_test_cuda.py
@@ -47,7 +47,7 @@ output_t custom_kernel(input_t data)
     cout, score = run_cuda_script(cu_eval, ref.read_text(), sub, arch=None)
     assert score == 0
     assert "Command '['./eval.out']' returned non-zero exit status 3." in cout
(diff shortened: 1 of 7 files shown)


scripts/ci_test_cuda.py
alexzhang13, line 1: I wonder if there's a nicer way to deal with this if we change eval.cu but since we don't do it frequently I guess it doesn't really matter lol

ngc92: I guess you could do a regex match with a wildcard for the line number, or just assert a smaller substring.

PR #127, opened 2025-01-13
Updated function signatures

ngc92:
This splits the run function into more easily testable portions, and makes it return a  structured result, instead of a string + score.

This will allow us to separate the handling of results (including calculating of scores, display or suppression of output etc) independently of the run script itself.


alexzhang13 (approved):
I like these changes a lot, probably want to port this Result struct to Python as well at some point (obviously not 1:1 bc of no distinct compile step, but generally it's just more readable than passing around strings).

PR #130, opened 2025-01-13
Unifying github and modal runners

ngc92:
this attempts to unify the code paths for github and modal, by reusing the same running scripts and discord output generation.
This is a major revision to how github runners are handled. Instead of individual files, we now submit all files packed up in a json object. Additionally, the python environment is provided in a separate `requirements` input.

Having  python deps be provided instead of hardcoded in the script is necessary, IMO, because  we probably want to pin pytorch/triton versions for a given leaderboard, but newer leaderboards should probably use the most recent pytorch. It also makes it possible for leaderboards to declare additional python dependencies in the future.

Similarly, we can now easily inject other header files to the cuda run.

One disadvantage of this change is that we now need to checkout the repository so we can use the helper scripts for the runner; meaning that the user code doesn't run in a pristine environment. For now, I think it's fine, but we probably should fix that in the future.
Another current shortcoming is that the singe-script use-case is no longer well supported. Again, I think we can leave that for a follow-up for now.

I've also extended the verify command to now test one of each, i.e., the subset of [modal, github] x [pytorch, cuda] x [nvidia, amd] which we currently support.


.github/workflows/nvidia_workflow.yml
alexzhang13, line 1: I quite like these changes. it's pretty clever and makes it way cleaner.

src/discord-cluster-manager/report.py
alexzhang13, line 1: This looks really clean on the UI side. We should have another discussion about exactly what information is sent to the user (e.g. do we provide them examples where it failed), but for this PR I am happy with this.

src/discord-cluster-manager/modal_runner_archs.py
alexzhang13, line 1: @ngc92 If you know of a cleaner way to set up these Modal runners for multiple GPUs than what we have now, that would be great as well. OTW this is currently functional and working, albeit quite messy.

src/discord-cluster-manager/leaderboard_eval.py
alexzhang13, line 1: Is there a way we can move Image requirements (e.g. TK / CUTLASS) like for Modal to be something like this as well at some point? Not sure exactly how this would go down but I kind of like moving all of the setup functionality for runners here. The only issue is that it's modal specific.

alexzhang13 (approved):
Looks good! All tested as well. The new threads are also much better and cleaner.

src/discord-cluster-manager/leaderboard_eval.py
ngc92: not sure, modal starts up much faster than github in part because we're pre-building everything, instead of running pip install for every submission.

PR #131, opened 2025-01-13
update pytorch script runner to use pipe

ngc92:
Updates the python interface to match  the recent cuda changes.
Turns out `run_cuda_program` actually doesn't do anything cuda specific, so we can instead reuse this piece of code.

In the current state, standalone python (and also cuda, I think) scripts are somewhat broken, as they never use the special pipe to communicate their success status, so these runs will now always be considered failed.


scripts/ci_test_cuda.py
alexzhang13, line 1: I would just recommend using Enums and maybe commenting about why we chose those run codes (because they're available lol) so this isn't confusing in the future. Esp if we iterate on this later.

scripts/ci_test_python.py
alexzhang13, line 1: Update exit codes as enums. (lines 24, 41, 51)

src/discord-cluster-manager/eval.cu
alexzhang13, line 1: Even though this is in the CUDA eval file, ideally we also make it clear here what these exit codes mean (I know it's kind of meh because we're basically defining the same Enum twice, once in the bot and once in here, but it'll help in the long run).

src/discord-cluster-manager/eval.py
alexzhang13, line 1: Same thing as .cu eval, you might have to create a third copy of the Enum here too... Because we ideally don't want to be importing from the bot code.

src/discord-cluster-manager/eval.cu
alexzhang13: Re: our discussion, no need to do this for the eval file.

src/discord-cluster-manager/eval.py
alexzhang13: Re: our discussion, no need to do this for the eval file.

alexzhang13 (approved):
All working and tested locally. Thanks!

PR #132, opened 2025-01-14
Allow arbitrary sets of files to be used in python and cuda runner code

ngc92:
For cuda, we specify a set of source files (to compile and link) and header files (just create them), for python a dict of python files and which of these should be executed as the main file.

This will go nicely together with the changes in #130 , and enable us to, e.g., useful utilities as common files. It also enables, on the cuda side, to compile user and tester code separately to prevent any include-based hacks to the tester.


.github/workflows/runner.py
alexzhang13, line 1: I just had to update the runner to handle the new type of file paylod

alexzhang13 (approved):

PR #134, opened 2025-01-14
generate_input receives a random seed argument

ngc92:
adds a seed argument to data generation functions.
**NOTE**  This requires an update to existing leaderboard reference files
Here, I've only updated the identity test so far.

### DIFF
--- a/examples/identity_cuda/reference.cuh
+++ b/examples/identity_cuda/reference.cuh
@@ -6,6 +6,7 @@
 #include <cstdlib>
 #include <cmath>
 #include <array>
+#include <random>
 #include <iostream>
 
 #define N_SIZES 10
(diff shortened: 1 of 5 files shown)


alexzhang13:
@ngc92  Just one Q about this, in our actual LB evals, how do we enforce a hidden seed that isn't visible to the user?

alexzhang13:
> **NOTE** This requires an update to existing leaderboard reference files

Can we set a default seed on the eval side? Or do you think this is dangerous?

alexzhang13 (approved):
I'm happy to merge this because the LBs on our server right now aren't really being used. We can just update the example kernels and the reference code once this is merged, just ping me.

ngc92:
I guess the question is whether we should merge this to main, or try to gather more breaking changes and  merge them all at once.

alexzhang13:
@ngc92 I think it's fine to merge in, at this point no one is really iterating on the LBs on the submission side so breaking changes are ok.

PR #135, opened 2025-01-16
further unification of runner code

ngc92:
Trying to align the modal and github runners so that only the absolutely necessary differences remain.


src/discord-cluster-manager/modal_runner_archs.py
alexzhang13, line 1: Super clean -- @ngc92 what does `serialized=True` do?

src/discord-cluster-manager/utils.py
alexzhang13, line 1: In the future I'm assuming you're going to make the reference content like a List or something right? If we want to support some kind of multi-file thing.

alexzhang13 (approved):
Love these changes, it's all coming together now -- everything tested locally is running so we should be good to go. Merge whenever @ngc92

src/discord-cluster-manager/modal_runner_archs.py
ngc92: Serialized means that we're somehow using python bytecode here, I  think.
If you don't set this, it tries to look up the function by name, and then complains
> AttributeError: module 'modal_runner' has no attribute 'run_cuda_script_t4'

src/discord-cluster-manager/utils.py
ngc92: a dict[str, str] file_name -> file_content, I think.

PR #136, opened 2025-01-17
Add a persistent Leaderboard Channel

alexzhang13:
## Description

Add a channel that just displays the leaderboard. Had to also split up cog functionality from forming the leaderboard to get this to work (you can't just call slash commands internally).

The read-only channel looks like so: It just repeatedly edits the same message (or adds an initial one) with an updated LB.

Screenshot 2025-01-17 at 3 51 41 PM

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/src/discord-cluster-manager/bot.py
+++ b/src/discord-cluster-manager/bot.py
@@ -97,6 +97,7 @@ class ClusterBot(commands.Bot):
         forum_channel = None
         submission_channel = None
         general_channel = None
+        leaderboard_channel = None
         for channel in category.channels:
             if channel.name == "central" and isinstance(channel, discord.ForumChannel):
                 forum_channel = channel
@@ -104,6 +105,8 @@ class ClusterBot(commands.Bot):
                 submission_channel = channel
             elif channel.name == "general" and isinstance(channel, discord.TextChannel):
                 general_channel = channel
+            elif channel.name == "active-leaderboards" and isinstance(channel, discord.TextChannel):
+                leaderboard_channel = channel
 
         if not forum_channel:
             forum_channel = await category.create_forum(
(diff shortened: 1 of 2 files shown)


ngc92:
instead of the loop, we could also just update whenever a new entry gets inserted into the database?

alexzhang13:
> instead of the loop, we could also just update whenever a new entry gets inserted into the database?

I don't mind this, I think both are pretty simple / won't affect the bot that much. The only nice thing is that we don't have to tie this to database logic, as it kind of just lives on its own in the LB class.

S1ro1:
> instead of the loop, we could also just update whenever a new entry gets inserted into the database?

Advantage of this is that we get the update instantly, however disadvantage is the decoupling of code. However, as this is quite user-facing functionality, I'd go with this suggestion @alexzhang13. I totally agree that from a standpoint of code design it's bit worse, but the functionality is "better".

alexzhang13:
> > instead of the loop, we could also just update whenever a new entry gets inserted into the database?
> 
> 
> 
> Advantage of this is that we get the update instantly, however disadvantage is the decoupling of code. However, as this is quite user-facing functionality, I'd go with this suggestion @alexzhang13. I totally agree that from a standpoint of code design it's bit worse, but the functionality is "better".

@S1ro1 Hm but there's a few things here that I'm not a fan of:

1. Our code base is changing quite rapidly. I could place these function calls in the LB cog class for submissions (I don't like this because we keep changing the cogs, and for good reason) or the DB class (I really don't like this because the cog functionality shouldn't be in the DB class at all) but it's highly likely we'll want to change where / how it's called in the near future if we do this. 

2. When we eventually put this into GPU MODE, I anticipate a decently high number of submissions per minute. This is fine, and in most cases unless there is a new leader often we won't even have to update the leaderboard, but I don't like the idea of constantly editing the message per submission if there are a lot.

3. So I think if we eventually do want to couple updating with the submission logic we should wait until we're happy with the code structure. We could just use this function for now and re-factor it when that time comes.

S1ro1 (approved):

alexzhang13:
Yeah I'm fine with his suggestions too I'd just prefer to wait until he's done refactoring before we add more logic to the submission flow.

PR #146, opened 2025-01-21
WIP: Tests and Benchmarks

ngc92:
Massive refactoring of our cuda eval template:
Now supports individual test cases and benchmarks, with corresponding commands.
At this point, the code is absolutely unpolished, but it serves to indicate what I'd like to achieve.

Still missing:
* [ ] python support
* [ ] test/benchmark timeouts
* [ ] actually randomizing the seed
* [ ] sanity check that secret run and public run are similar

Long shots.
* Additional "profile" run: Run a single benchmark under the profiler.

Q:
* Allow test/benchmark submissions via API, leaderboard only via discord?
* Run public/private leaderboard submissions in sequence or in parallel?
* Database changes to accomodate public and private runs?


alexzhang13 (approved):
Tested locally, all bugs @S1ro1 and I found were fixed, we'll most other TODOs in a different PR.

PR #149, opened 2025-01-22
Add MI300X Runners to AMD workflow.

saienduri:
## Description

This commit adds mi300 nodes as an option to the current amd workflow.

I tested a couple times on the AMD workflow here (https://github.com/gpu-mode/discord-cluster-manager/actions/runs/12902691574)

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.

### DIFF
--- a/.github/workflows/amd_workflow.yml
+++ b/.github/workflows/amd_workflow.yml
@@ -68,6 +68,6 @@ jobs:
       uses: actions/upload-artifact@v4
       if: always()
       with:
-        name: run-result-${{ matrix.name }}
+        name: run-result
         path: |
           result.json
--- a/.github/workflows/amd_workflow.yml
+++ b/.github/workflows/amd_workflow.yml
@@ -13,10 +13,18 @@ on:
 
 jobs:
   run:
-    runs-on: [amdgpu-mi250-x86-64]
+    runs-on: ${{ matrix.runs-on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: mi300
+            runs-on: amdgpu-mi300-x86-64
+            venv-dir: ${{ github.workspace }}
+          - name: mi250
+            runs-on: amdgpu-mi300-x86-64
+            venv-dir: /groups/aig_sharks/pytorch_venv
     timeout-minutes: 10
-    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/amd_workflow.yml
```diff
@@ -13,10 +13,18 @@ on:
 
 jobs:
   run:
-    runs-on: [amdgpu-mi250-x86-64]
+    runs-on: ${{ matrix.runs-on }}
+    strategy:
```
ngc92, line 17: I  don't think this is the way we want to handle multiple runners; afaics, this well _always_ run the given code sample on both types of AMD cards. Instead, the card should be selectable by an  input argument that is given to the  job.

Both runners should be registered in src/discord-cluster-manager/consts.py, and 
https://github.com/gpu-mode/discord-cluster-manager/blob/main/src/discord-cluster-manager/cogs/github_cog.py#L35-L40
needs to be updated

.github/workflows/amd_workflow.yml
```diff
@@ -48,13 +55,14 @@ jobs:
...
+        source ${{ matrix.venv-dir }}/bin/activate
         python3 .github/workflows/runner.py
         cat result.json  # Debug: show output
 
     - name: Upload training artifacts
       uses: actions/upload-artifact@v4
       if: always()
       with:
-        name: run-result
+        name: run-result-${{ matrix.name }}
```
ngc92, line 66: this name needs to be unchanged so the bot knows how to read the results.

msaroufim:
To merge this PR faster I'm thinking what might be easiest is just moving to only using MI300 - wdyt @saienduri

saienduri:
Thanks for the signal @msaroufim! I updated the PR, so that it should work with card selection. Does it look good to y'all?
Here's an example run I trigged in github: https://github.com/gpu-mode/discord-cluster-manager/actions/runs/13350719364

.github/workflows/amd_workflow.yml
saienduri: Updated github_cog but I don't think we need to touch consts.py

msaroufim (approved):

PR #152, opened 2025-01-25
Make run_cuda_script accept more compile flags arguments

ngc92:
## Description
* Adds additional arguments for cuda compiles that allow specifying link libraries, defines, and generic flags.
* Additional verification and unit tests.

The additional checks that supplied include directories actually exist required to move the modal-specific thunderkittens include into the modal cog.

### DIFF
--- a/scripts/ci_test_cuda.py
+++ b/scripts/ci_test_cuda.py
@@ -1,6 +1,9 @@
 import os
 import sys
+import tempfile
 from pathlib import Path
+
+import pytest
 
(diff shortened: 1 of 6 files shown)


scripts/ci_test_cuda.py
alexzhang13, line 1: I'm going to trust that these tests are sufficient, but they look fine to me.

alexzhang13 (approved):
Tested with some really hacky stuff on my end, but I wasn't able to find any issues. I like these changes :)

PR #169, opened 2025-02-19
Update of result structures and eval runs

ngc92:
Update evalulation runs:
* Attach a potential compile result to _each_ run (we might want to compile tests without `-DNDEBUG`, for example; not part of this PR, though)
* record start and end time of runs
* have task specific timeouts for test, benchmark, and ranked runs, with reasonable defaults (i hope; maybe they should be even lower?)
* restructuring of eval code

Testing these changes requires making a deploy to modal so the new test script is there.

### DIFF
--- a/src/discord-cluster-manager/run_eval.py
+++ b/src/discord-cluster-manager/run_eval.py
@@ -1,4 +1,6 @@
 import dataclasses
+import datetime
+import functools
 import os
 import shlex
 import subprocess
@@ -6,7 +8,7 @@ import tempfile
 import time
 from pathlib import Path
 from types import NoneType
-from typing import Optional, Union
+from typing import Callable, Optional, Union
 
 from consts import CUDA_FLAGS, ExitCode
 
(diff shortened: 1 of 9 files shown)


src/discord-cluster-manager/run_eval.py
```diff
@@ -38,15 +40,24 @@ class RunResult:
...
 @dataclasses.dataclass
 class FullResult:
     # fmt: off
     success: bool                  # did the runner (github/modal) execute successfully
     error: str                     # if not success, an error message
-    compile: CompileResult | None  # results of compilation
     # results of running. There can be multiple runs in one submission, using separate
     # 'test' and 'benchmark' keys, for example
-    runs: dict[str, RunResult] = dataclasses.field(default_factory=dict)
+    runs: dict[str, EvalResult] = dataclasses.field(default_factory=dict)
```
S1ro1, line 60: Maybe try to think of a better name for this? It is slightly misleading and hard to read, ie. when used as `test_run.run`. Maybe `run_results`?

PR #170, opened 2025-02-19
msaroufim/test amd

msaroufim:
AMD requirements are now failing with

```
 ERROR: Could not find a version that satisfies the requirement torch==2.6.0.dev20241023+rocm6.2 (from versions: 2.6.0.dev20241122+rocm6.2)
ERROR: No matching distribution found for torch==2.6.0.dev20241023+rocm6.2
```

We are using an "old" nightly version which was likely recently purged from the PyTorch index, instead we'll use the stable release for 6.2.4 and remove the explicit installation of triton since PyTorch also packages triton when making releases

This fixes that issue but now there's another real looking failure in the leaderboard code https://github.com/gpu-mode/discord-cluster-manager/actions/runs/13422933345/job/37499644756

### DIFF
--- a/src/discord-cluster-manager/consts.py
+++ b/src/discord-cluster-manager/consts.py
@@ -139,7 +139,5 @@ triton
 """
 
 AMD_REQUIREMENTS = """
---index-url https://download.pytorch.org/whl/nightly/rocm6.2
-pytorch-triton-rocm==3.1.0+cf34004b8a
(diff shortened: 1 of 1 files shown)


msaroufim:
CLosing since this will get merged with @ngc92's changes in another PR

ngc92:
let's keep this independently, I rebased the other  PR only for testing

ngc92 (approved):

PR #171, opened 2025-02-20
Database update to record submissions

ngc92:
This is a first attempt at reworking our database to enable detailed recording of submissions.
It split submissions into 3 tables:
1) code files. Here we store the actually submitted code. We have a hash column that allows us to quickly detect if the same code is submitted again, and in this case we just don't add it (e.g., multiple commands run with the same code, e.g., multiple benchmarks or multiple runs explicitly specifying different GPUs, or everyone testing our example script :)), no need to spam our DB with redundant data.
2) A table for an individual "submission", that is, one invocation of a bot command. This is associated with a code, and adds information about submission time, leaderboard and user. We also record a filename, and have a special flag "done" that starts out false and gets set to true once all jobs have finished. If we find lots of not-done submissions in our DB, we know something is wonky in the bot...
3) Actual runs. Executing the submission on a specific GPU, with a specific set of tests or benchmarks. We record start and end time of the actual run (might come in handy if we want to see if particular times of day have faster/slower GPUs, or if  we want to keep track of the average delay between submission and a job starting), which  mode (test, benchmark, profile, etc), which runner (this is essentially gpu type, but if we want to differentiate modal_h100 and github_h100, I think runner might be the better term); we  also keep all the other metadata (compilation result, running stdout/stderr, etc) in jsonb objects. Having _all_ runs, and all their start/end times in the  DB also enables us to figure out if particular users are overusing our hardware resources.

This is deliberately proactive in gathering as much data as possible, trying to minimize the information we throw away. For the leaderboard itself, we don't really need to keep track of submissions that don't have an associated score. Still, I'd rather record all submissions, and at some point start pruning the DB and throw away (user-invisible) things that aren't interesting. But first, let's see how much data this actually produces.

There is a new show-stats admin command, that quickly gives us an overview over the database:
```
num_unique_codes = 3
num_submissions = 5
sub_waiting = 0
num_users = 1
sub_last_day = 5
num_runs = 13
runs_passed = 13
runs_scored = 4
runs_secret = 6
num_run.L4 = 1
runs_passed.L4 = 1
runs_scored.L4 = 0
runs_secret.L4 = 0
num_run.T4 = 12
runs_passed.T4 = 12
runs_scored.T4 = 4
runs_secret.T4 = 6
total.xxxxxxxxxx = 0:00:00.019583
```
the last is my user-id, showing that I've used 19ms of runner time (excluding any setup/teardown, etc) in the last 24h.
this would show us the users with the highest runner usage.

### DIFF
--- a/src/discord-cluster-manager/cogs/admin_cog.py
+++ b/src/discord-cluster-manager/cogs/admin_cog.py
@@ -76,6 +76,10 @@ class AdminCog(commands.Cog):
             name="update-problems", description="Reload all problem definitions"
         )(self.update_problems)
 
+        self.show_bot_stats = bot.admin_group.command(
+            name="show-stats", description="Show stats for the bot"
+        )(self.show_bot_stats)
+
     # --------------------------------------------------------------------------
     # |                           HELPER FUNCTIONS                              |
     # --------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


S1ro1 (approved):
lgtm

PR #172, opened 2025-02-20
error handling and logging

ngc92:
some logging improvements
created a dedicated Exception subclass that we can use to communicate error messages across call boundaries. We don't want to use  the original exceptions, because we cannot actually display them to the user as we just don't know what kind of info might be visible in the exception object. For `KernelBotError`, the idea is that this indicates that the message of the exception object is intended specifically to be displayed to the user.

This avoids some cases where our db functions return strings only for the purpose of allowing error handling, and makes it harder to accidentally ignore such errors.

Additionally, contains some clean-up fixes, such as moving `resync` to the `admin` commands.

### DIFF
--- a/src/discord-cluster-manager/leaderboard_db.py
+++ b/src/discord-cluster-manager/leaderboard_db.py
@@ -16,7 +16,7 @@ from env import (
 )
 from run_eval import CompileResult, RunResult
 from task import LeaderboardTask
-from utils import LeaderboardItem, LeaderboardRankedEntry, LRUCache, setup_logging
+from utils import KernelBotError, LeaderboardItem, LeaderboardRankedEntry, LRUCache, setup_logging
 
 leaderboard_name_cache = LRUCache(max_size=512)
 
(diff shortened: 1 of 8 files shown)


src/discord-cluster-manager/utils.py
```diff
@@ -107,17 +134,6 @@ async def send_logs(thread: discord.Thread, logs: str) -> None:
         await thread.send(f"```\n{chunk_text}\n```")
 
 
-def extract_score(score_str: str) -> float:
```
msaroufim, line 110: do we no longer need this?

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -68,8 +69,8 @@ def connect(self) -> bool:
             )
             self.cursor = self.connection.cursor()
             return True
-        except Error as e:
-            print(f"Error connecting to PostgreSQL: {e}")
+        except psycopg2.Error as e:
+            logger.exception("Error connecting to PostgreSQL", exc_info=e)
```
msaroufim, line 73: presumably this is on by default since python logging is warn level

msaroufim (approved):
nice!

src/discord-cluster-manager/utils.py
ngc92: no, haven't needed this for a long time. this was for parsing stdout to try and  find a substring that looked like it wanted to report a score

PR #173, opened 2025-02-21
Leaderboard Website

msaroufim:
## Technical decisions

The way this PR works is 

On a nightly basis or push to main, we manually generate a file called `leaderboard.html` which is generated by the script `scripts/update-leaderboard.py`

This file gets created in our regular doc build workflow, and is checked into the repo before the final doc build can be made

It's all a very hacky way of turning a static site generator like Github Pages into a dynamic scheduler.

## Design decisions

In general we want to show a hierarchy of
* GPU type
* Leaderboard problem
* User submisssion

The intent here is that some leaderboards are more interesting than others, example H100 and the list isn't massive to navigate

There are probably tons of other UX changes that can be considered
1. Which season we're part of 
2. Collapsable GPU type
3. Filter on problem

## Local test

```
npm install 
npm run start
```

Which gives the below screenshot

Screenshot 2025-02-20 at 6 43 10 PM

## Remote Tests

I split up the github action so we always build locally but only deploy on main. Build is green so we should be good 

## Open issues
1. Our DB doesn't have discord usernames, only discord IDs, this is fine and I can fix in a followup PR 
5. leaderboads in staging server right now only have T4 numbers and all submissions are crashing so can't populate with more data. But this is correct in that this is all the data available in our DB after I ran SQL queries on heroku


msaroufim:
Alex checked this so over DM so just gonna merge

PR #184, opened 2025-02-23
better verify

ngc92:
Updates to `verify-task` : 
 * can decide to run only test/benchmark/ranked or all
 * looks for things to run in `solutions/{correct,wrong}` subfolders
 * automatically verifies that `correct`  solutions get `passed`  status, and `wrong` solutions don't


S1ro1 (approved):
lgtm

PR #191, opened 2025-02-23
Timeout related fixes

ngc92:
* fix setting correct end_time for python jobs
* add a "compile" step to python to pre-populate `load_inline` cache


S1ro1 (approved):
lgtm 🚀

PR #205, opened 2025-03-04
WIP: admin delete submission

ngc92:
Adds an admin command to delete a submission of a given ID, taking care to delete the corresponding runs first.

### DIFF
--- a/src/discord-cluster-manager/leaderboard_db.py
+++ b/src/discord-cluster-manager/leaderboard_db.py
@@ -591,6 +591,30 @@ class LeaderboardDB:
 
         return result
 
+    def delete_submission(self, submission_id: int):
+        try:
+            # first, the runs
+            query = """
+                    DELETE FROM leaderboard.runs
+                    WHERE submission_id = %s
+                    """
+            self.cursor.execute(query, (submission_id,))
+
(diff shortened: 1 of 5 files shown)


src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -591,6 +591,30 @@ def _generate_stats(self):
...
         return result
 
+    def delete_submission(self, submission_id: int):
+        try:
+            # first, the runs
+            query = """
+                    DELETE FROM leaderboard.runs
+                    WHERE submission_id = %s
+                    """
+            self.cursor.execute(query, (submission_id,))
```
b9r5, line 601: This makes me wonder if we should have made the foreign key from runs to submission be `on delete cascade`.

```
> \d leaderboard.runs
                                            Table "leaderboard.runs"
    Column     |           Type           | Collation | Nullable |                   Default                    
---------------+--------------------------+-----------+----------+----------------------------------------------
...
 submission_id | integer                  |           | not null | 
...
Foreign-key constraints:
    "runs_submission_id_fkey" FOREIGN KEY (submission_id) REFERENCES leaderboard.submission(id)
```

PR #207, opened 2025-03-05
Record system information

ngc92:
## Description

An attempt to record more information about the system we're running on. If we'd done that from the beginning, we would've been able to identify the submissions that were affected by our unspecific modal runner selection.

This implementation imports pytorch to get the GPU name from python, which would be required now even for pure cuda submissions.


msaroufim (approved):

S1ro1:
lgtm from me as well, tested out

PR #224, opened 2025-03-22
Keeping track of forum thread

ngc92:
This PR keeps track of the id of the forum thread associated with a leaderboard.
In itself, this isn't very useful yet, but in the future, it lets us communicate more easily in a leaderboard-specific channel.


S1ro1 (approved):
Like the idea, lgtm. Don't have a capacity to test locally rn, so merge if you're sure @ngc92

PR #228, opened 2025-03-29
Enable secret seeds

ngc92:
Enable (and use) secret seeds in the default python runner (`examples/eval.py`).
Secret seeds are not used in `eval.cu` yet.

Requires a modal redeploy.
Because `eval.py` is baked into the task definitions, none of the existing tasks will benefit from this. We'll need to figure out how to update existing tasks separately.

### DIFF
--- a/examples/eval.py
+++ b/examples/eval.py
@@ -5,7 +5,7 @@ import os
 import sys
 import math
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
 
 import torch.cuda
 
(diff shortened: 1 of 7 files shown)


examples/eval.py
```diff
@@ -236,13 +252,12 @@ def main():
         return 2
 
     mode = sys.argv[1]
-    tests = get_test_cases(sys.argv[2])
+    seed = os.getenv("POPCORN_SEED")
```
S1ro1, line 255: A small nit:

```suggestion
    seed = int(os.getenv("POPCORN_SEED", 42))
    set_seed(seed)
    tests = get_test_cases(sys.argv[2], seed)
```

S1ro1 (approved):
Except of this, LGTM!

examples/eval.py
ngc92: not sure about these changes:
1) I really dislike set_seed, it adjusts global state, and if you were to actually use it you couldn't run a subset of the tests; you would have to run them all, and in the same order.
2) I explicitly want to have the None value here, which tells `get_test_cases` to use the seeds from the `yml` file as-is.

We could put an unconditional `set_seed(42)` call here, as an insurance policy in case a test unintentionally relies on a source of randomness not controlled by the explcitly-passed `seed` argument.

Maybe we should actually add a new bot commend "check-task", that runs the test case generators in randomized order an checks that they're reproducible?

S1ro1: Ahh, makes sense, anyway it's approved so merge anytime @ngc92. Probably let's do it before #230 as one will need to update migrations to match the previous one.

PR #230, opened 2025-04-03
Feat: Authenticate from CLI

S1ro1:
Enables authentication via discord oauth -> we set redirect url to our `/auth/cli` (we figure out a proper name later), each cli instance gets a unique `cli_id` generated which maps user to cli. This cli gets encoded into `state` of the oauth2 redirect and sent to the server's auth url (/auth/cli).
Server then calls discord api to get user data, and inserts it into the database table, mapping the cli_id to the user_id.

To ban a user, we just ban based on discord username/user_id, this then doesn't allow them to auth via cli either. (NOT IMPLEMENTED YET)

This will then need a followup pr making each route also accept cli_id in the params.

Also possible update is to only use global discord names for the leaderboard, we should unify that (will tackle in another PR as well)

Makes heavy use of extra env vars (I'm not a fan of this), but they don't touch existing code at all so it's w.e.

### DIFF
--- a/src/discord-cluster-manager/api/main.py
+++ b/src/discord-cluster-manager/api/main.py
@@ -119,14 +119,14 @@ async def cli_auth(code: str, state: str = None):
 
     try:
         cli_id = base64.b64decode(state).decode("utf-8")
-    except Exception as e:
-        raise HTTPException(status_code=400, detail=f"Invalid state parameter: {str(e)}") from e
+    except Exception:
+        raise HTTPException(status_code=400, detail="Invalid state parameter") from None
 
     with bot_instance.leaderboard_db as db:
         try:
             db.create_user_from_cli(user_id, user_name, cli_id)
-        except Exception as e:
-            raise HTTPException(status_code=400, detail="Failed to create user") from e
+        except Exception:
+            raise HTTPException(status_code=400, detail="Failed to create user") from None
 
     return {"status": "success", "user_id": user_id, "cli_id": cli_id, "user_name": user_name}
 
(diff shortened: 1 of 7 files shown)


src/discord-cluster-manager/api/main.py
```diff
@@ -53,6 +57,83 @@ async def update(self, message: str):
...
+
+    user_response = requests.get(user_url, headers=headers)
+    if user_response.status_code != 200:
+        raise HTTPException(status_code=401, detail="Failed to retrieve user information")
+
+    user_json = user_response.json()
+    user_id = user_json.get("id")
+    user_name = user_json.get("username")
+
+    if not state:
```
ngc92, line 120: state is already checked at the top of the function

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -743,6 +743,26 @@ def get_leaderboard_submission_count(
...
+                ON CONFLICT (id) DO UPDATE
+                SET user_name = %s, cli_id = %s
+                """,
+                (user_id, user_name, cli_id, user_name, cli_id),
+            )
+            self.connection.commit()
+        except psycopg2.Error as e:
+            self.connection.rollback()
+            logger.exception("Could not create/update user %s from CLI.", user_id, exc_info=e)
+            raise KernelBotError(f"Could not create/update user {user_id} from CLI.") from e
```
ngc92, line 764: KernelBotError is intended to convey error messages to the end user, not exposing any internal data (that we'd log with logger). 
AFAICS, this just triggers `raise HTTPException(status_code=400, detail="Failed to create user")`, so just re-raising the psycopg2 error would be fine here?

ngc92 (approved):

PR #248, opened 2025-04-15
Post-launch fixes

ngc92:
Bunch of small fixes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=postgres
-      - POSTGRES_DB=mydb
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=postgres
-      - POSTGRES_DB=mydb
+      - POSTGRES_DB=clusterdev
```
S1ro1, line 10: This docker is only used when you run it locally, so I would not change it as it could change your local stuff.

S1ro1 (approved):

docker-compose.yml
ngc92: except that this is the name of the db that we put in the readme file.

PR #249, opened 2025-04-15
Feat: streaming response

S1ro1:
WIP: have to recheck when I have fresh mind, but should avoid the timeouts from heroku by utilizing streaming responses.

### DIFF
--- a/src/discord-cluster-manager/api/main.py
+++ b/src/discord-cluster-manager/api/main.py
@@ -1,5 +1,6 @@
 import asyncio
 import base64
+import datetime
 import json
 import os
 import time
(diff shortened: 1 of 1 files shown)


src/discord-cluster-manager/api/main.py
```diff
@@ -213,15 +222,75 @@ async def cli_auth(auth_provider: str, code: str, state: str):  # noqa: C901
...
+        )
+
+        while not task.done():
+            elapsed_time = time.time() - start_time
+            yield f"event: status\ndata: {json.dumps({'status': 'processing',
+                                                      'elapsed_time': round(elapsed_time, 2)}
+                                                      ,default=json_serializer)}\n\n"
+
+            try:
+                await asyncio.wait_for(asyncio.shield(task), timeout=15.0)
```
ngc92, line 247: a comment here would be nice. shield makes it so that the timeout does not affect the task itself?

ngc92 (approved):

PR #280, opened 2025-05-12
Fix Timeout Bug

PaliC:
## Description

This pr makes it such that we actually use the specified timeouts from github when running a submission. This logic was missing before. There's also a bunch of refactoring changes to make the linter happy.

Testing:
There is now a snippet in the logs that looks like 
```
2025-05-14 17:26:57 - utils - INFO - Waiting for workflow to complete... (timeout: 3 minutes)
2025-05-14 17:26:57 - utils - INFO - Workflow dispatch successful. Waiting 5s for the run to appear..
```
which indicates that we are passing a non default value. The associated workflow is https://github.com/gpu-mode/discord-cluster-manager/actions/runs/15033809006 (I believe the failure is noisy)

### DIFF
--- a/src/discord-cluster-manager/launchers/github.py
+++ b/src/discord-cluster-manager/launchers/github.py
@@ -13,6 +13,7 @@ from consts import (
     DEFAULT_GITHUB_TIMEOUT_MINUTES,
     GPU,
     NVIDIA_REQUIREMENTS,
+    TIMEOUT_BUFFER_MINUTES,
     GitHubGPU,
     SubmissionMode,
 )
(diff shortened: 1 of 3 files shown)


ngc92:
it seems you have moved the `GitHubLauncher` class around, which makes it a bit difficult to see the actually relevant parts in the diff.

src/discord-cluster-manager/launchers/github.py
```diff
@@ -70,7 +87,13 @@ async def run_submission(
 
         await status.push("⏳ Waiting for workflow to start...")
         logger.info("Waiting for workflow to start...")
-        await run.wait_for_completion(lambda x: self.wait_callback(x, status))
+
+        timeout = get_timeout(config)
```
ngc92, line 91: we should add one or two minutes to account for github setting up the container etc.

PaliC: done

ngc92 (approved):
LGTM

PR #283, opened 2025-05-20
add lb option to show-stats

PaliC:
This is a pretty simple pr that adds an lb option to show-stats. Right now it just shows average runtime as an extra stat, but I'll add a baseline time in a future pr (this is involved enough I wanted to  break it out)

Here's a pic of stats. First one is the default and the second is for a single leaderboard.

Screenshot 2025-05-23 at 12 30 20 PM

### DIFF
--- a/.github/workflows/nvidia_workflow.yml
+++ b/.github/workflows/nvidia_workflow.yml
@@ -28,6 +28,8 @@ jobs:
     - name: Create input files
       shell: bash
       run: |
+        # install jq
+        apt update && apt install -y jq
         # Extract the payload content without printing it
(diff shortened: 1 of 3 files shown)


msaroufim:
So main reason I'm pausing on this is because I think we should probably just create a table with admin stats and then we can have significantly simpler queries for both the per leaderboard stats and the per day stats

PaliC:
@msaroufim Sorry didn't see this for a hot sec. What are the admin stats here? Would they be delay? The rest seem like stats we should store in a leaderboard table regardless.

If you're referring to the baseline, I can move that to another table (though the PR I have puts it in the leaderboard table as well)

PR #286, opened 2025-05-30
Feat: enable new modal gpus 🚀

S1ro1:
Enable H200 + B200 for Modal

### DIFF
--- a/src/discord-cluster-manager/consts.py
+++ b/src/discord-cluster-manager/consts.py
@@ -29,6 +29,7 @@ class ModalGPU(Enum):
     A100 = "A100"
     H100 = "H100"
     B200 = "B200"
+    H200 = "H200"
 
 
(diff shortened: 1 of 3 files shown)


src/discord-cluster-manager/consts.py
```diff
@@ -115,6 +116,7 @@ class RankCriterion(Enum):
     "L4": "80",
     "A100": "80",
     "H100": "90a",
+    "H200": "90a",
```
msaroufim, line 119: should we enable H200 tho?

S1ro1: Don't have a strong opinion, probably not too interesting GPU so idc if we remove.

msaroufim: yeah'd rather we remove

PR #288, opened 2025-05-31
service refactor

msaroufim:
- **Feat: backend db service**
- **Feat: show/list works**


S1ro1:
This is pretty much unmergable, should serve only as a general overview of how the refactor could look. Claude has cooked way too much, implementation wise. Design-wise, this is what I'd aim for.

PR #298, opened 2025-06-13
[ez] install jq in nvidia workflow

PaliC:
jq is not in the docker image during local development so this pr installs it explictly


ngc92:
I think copilot is right here? Was this maybe added by another PR in the meantime?

PR #299, opened 2025-06-13
[ez] update references so they run locally

PaliC:


ngc92 (approved):
the nitpicks by copilot seem correct, but not critical

PR #300, opened 2025-06-15
Disentangling discord

ngc92:
In preparation for further cleanups, this PR tries to disentangle our files so there is less (direct and indirect) importing of discord happening.

* db is not independent from env
* db helper typedefs are now in db file (not 100% sure about that one, but imo a better place than "utils")
* instead of a global, the db stores the LRU cache as a member
* moved all utilities that interact with discord into a separate file

This should lay the groundwork so that we can make a libkernelbot python package that is completely independent from discord, and gets imported by the bot. Once we have this, we can start moving more functionality into this lib (looking at you, `reports.py`)

The PR contains a bunch of changes, but each individual commit should be fairly straightforward.

### DIFF
--- a/src/discord-cluster-manager/leaderboard_db.py
+++ b/src/discord-cluster-manager/leaderboard_db.py
@@ -5,10 +5,6 @@ import logging
 from typing import List, Optional
 
 import psycopg2
-from env import (
-    DATABASE_URL,
-    DISABLE_SSL,
-)
 from run_eval import CompileResult, RunResult, SystemInfo
 from task import LeaderboardTask
 from utils import (
@@ -25,7 +21,9 @@ logger = setup_logging(__name__)
 
 
 class LeaderboardDB:
-    def __init__(self, host: str, database: str, user: str, password: str, port: str = "5432"):
+    def __init__(
+        self, host: str, database: str, user: str, password: str, port: str, url: str, ssl_mode: str
+    ):
         """Initialize database connection parameters"""
         self.connection_params = {
             "host": host,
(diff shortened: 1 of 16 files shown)


ngc92:
@S1ro1 bump :)

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -393,7 +389,7 @@ def get_leaderboard_gpu_types(self, leaderboard_name: str) -> List[str] | None:
         else:
             return None
 
-    def get_leaderboard(self, leaderboard_name: str) -> LeaderboardItem | None:
+    def get_leaderboard(self, leaderboard_name: str) -> "LeaderboardItem | None":
```
S1ro1, line 391: Shouldn't be "LeaderboardItem" | None. This probably doesn't change anything, but still

src/discord-cluster-manager/leaderboard_db.py
```diff
@@ -893,3 +889,52 @@ def validate_cli_id(self, cli_id: str) -> Optional[dict[str, str]]:
             self.connection.rollback()
             logger.exception("Error validating CLI ID %s", cli_id, exc_info=e)
             raise KernelBotError("Error validating CLI ID") from e
+
+
+class LeaderboardItem(TypedDict):
```
S1ro1, line 893: Nit: let's move these to the top of the file probably? Bit annoying to scroll through whole database to get to this IMO

src/discord-cluster-manager/leaderboard_db.py
ngc92: wouldn't this implicitly call operator| on a str object?

PR #301, opened 2025-06-17
add a backend class that handles db and submission triggering

ngc92:
This moves the remaining critical bits of submission logic out of SubmitCog, which now only exists to submit the legacy  "run" commands; don't think we've ever used them, maybe just nuke them to get rid of the complexity?

There are still lots of submission-related things scattered and duplicated all across the codebase, but if i didn't miss anything, this should make it possible to write a new/additional main file that enables running the API without the discord bot being required.

### DIFF
--- a/src/discord-cluster-manager/api/main.py
+++ b/src/discord-cluster-manager/api/main.py
@@ -227,7 +227,6 @@ async def cli_auth(auth_provider: str, code: str, state: str, db_context=Depends
 
 async def _stream_submission_response(
     submission_request: SubmissionRequest,
-    user_info: dict,
     submission_mode_enum: SubmissionMode,
     backend: KernelBackend,
 ):
@@ -237,7 +236,6 @@ async def _stream_submission_response(
         task = asyncio.create_task(
             _run_submission(
                 submission_request,
-                user_info,
                 submission_mode_enum,
                 backend,
             )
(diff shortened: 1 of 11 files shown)


S1ro1:
I'd be in favour of just nuking the whole SubmitCog now, rather than later.

src/discord-cluster-manager/api/main.py
```diff
@@ -50,13 +55,29 @@ async def simple_rate_limit():
...
+    """Database context manager with guaranteed error handling"""
+    if not backend_instance:
+        raise HTTPException(status_code=500, detail="Bot instance not initialized")
+
+    if not hasattr(backend_instance, "leaderboard_db"):
+        raise HTTPException(status_code=500, detail="Database not initialized")
+
+    with backend_instance.db as db:
+        if db is None:
+            raise HTTPException(status_code=500, detail="Database connection failed")
```
S1ro1, line 74: Let's remove these 2 error checks, IMO this never happens and it's just an AI slop that I left behind from when API was just a PoC

S1ro1: Maybe even the 1st check is useless, thinking of it. If it ever triggers there is just nothing for us to do.

S1ro1 (approved):
LGTM, again only small nits

PR #303, opened 2025-06-18
remove SCRIPT run mode

ngc92:
As discussed; we have never actually used the SCRIPT mode in the last 6 months or so; no reason to keep the complexity around for that. If we really need it, we can reimplement.
Right now, this removes then entire  SubmitCog, and a fair bit of branching in other code areas.


S1ro1 (approved):
LGTM. Love to see this gone

PR #305, opened 2025-06-19
Separate out templates into their own table

ngc92:
This change focuses the "task" column to be more specifically tailored towards the actual runner's task, removing metadata (that is, description and templates) that is not directly required to a separate column/table.

This can serve as a template ( :) ) for how to handle future ancillary data, such as milestones.


PaliC (approved):
Generally this is pretty intuitive imo. Only nit is that we just need to ensure we can tell what a LeaderBoardTask vs LeaderBoardDefinition is at a glance for the yaml file (though this can be done stylistically / with comments on the .yml file)

PR #319, opened 2025-07-27
Adding unit tests

ngc92:
This PR adds a bunch of unit tests for libkernelbot. While doing so, slight improvements (error checking) were made to submission.py.
This doesn't cover all the files in libkernelbot, but `consts.py` and `db_types.py` just define dataclasses/enums so there's not much to test there, and `leaderboard_db.py`, `backend.py` how much **unit** testing we can actually do there (i.e., is it even worth trying to mock the db, or should we just aim for some integration tests that actually start the postgres service and populate a real database?)

The tests for report.py are mostly claude generated, with some strong opinionated guidance to make them somewhat manageable.

### DIFF
--- a/.github/workflows/runner_ci.yml
+++ b/.github/workflows/runner_ci.yml
@@ -83,3 +83,11 @@ jobs:
     env:
       CUDA_VISIBLE_DEVICES: 0
 
+  unit-tests:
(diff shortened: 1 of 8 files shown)


msaroufim (approved):
Amazing

PR #320, opened 2025-07-28
Database tests

ngc92:
Uses docker-compose to start up a container with the db, and  run the yoyo migrations, so we can access a real database for our testing.
All of this is wrapped inside a pytest fixture, so for local testing you can just run pytest as normally. However, this means that pytest now takes a few seconds to start up, because it waits on the docker container being available.

While developing this, I've fixed a few minor and a major problem. Please take a **careful** look at the migration here, and check if it makes sense to you. 

Apart from the code file fix, this fixes a sign error when getting leaderboard names (for autocomple) and does some small cleanups/error handling improvements.

The code paths where the database returns an error are currently ignored in coverage reporting; while for some situations, we can trigger them easily (and in fact, some of the tests do), trying to make the tests comprehensive here doesn't really seem worth the effort. 
Furthermore, the user-id/user-name/CLI-user stuff doesn't get any tests yet. Probably needs a rework anyway.

### DIFF
--- a/src/migrations/20250728_01_Q3jso-fix-code-table.py
+++ b/src/migrations/20250728_01_Q3jso-fix-code-table.py
@@ -21,10 +21,13 @@ def convert_code_to_bytea(conn):
     records = cursor.fetchall()
 
     existing_codes = {}
+    num_duplicates = 0
+    num_encoded = 0
 
     for record_id, code_text in records:
         # broken with the old code
         if code_text.startswith("\\x"):
+            num_encoded += 1
             code_text = bytes.fromhex(code_text[2:]).decode("utf-8")
         code_bytes = code_text.encode("utf-8")
         # with the old broken code and experimentation, it is possible that we got some
@@ -35,6 +38,7 @@ def convert_code_to_bytea(conn):
                 (existing_codes[code_bytes], record_id),
             )
             cursor.execute("DELETE FROM leaderboard.code_files WHERE id = %s", (record_id,))
+            num_duplicates += 1
             continue
 
         existing_codes[code_bytes] = record_id
@@ -43,6 +47,9 @@ def convert_code_to_bytea(conn):
         cursor.execute(
             "UPDATE leaderboard.code_files SET code = %s WHERE id = %s", (code_bytes, record_id)
         )
+
+    print(f"Found and removed {num_duplicates} duplicates")
+    print(f"Reencoded {num_encoded} code submissions")
 
 
 def convert_bytea_to_text(conn):
--- a/src/migrations/20250728_01_Q3jso-fix-code-table.py
+++ b/src/migrations/20250728_01_Q3jso-fix-code-table.py
@@ -78,15 +78,16 @@ steps = [
     # run the conversion
     step(convert_code_to_bytea, convert_bytea_to_text),
     # clean up the table and reintroduce hashes
+    # ALTER TABLE leaderboard.code_files DROP COLUMN old_code;
+    # do this later, once we're confident that the migration works
     step(
         """
-       ALTER TABLE leaderboard.code_files DROP COLUMN old_code;
+       ALTER TABLE leaderboard.code_files ALTER COLUMN old_code DROP NOT NULL;
        ALTER TABLE leaderboard.code_files ADD COLUMN hash TEXT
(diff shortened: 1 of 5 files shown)


msaroufim:
@b9r5 we're really worried about getting this PR wrong, would love your feedback especially on the PR migration, we'd want to make sure that we get weird character encodings finally correct

src/migrations/20250728_01_Q3jso-fix-code-table.py
```diff
@@ -0,0 +1,93 @@
...
+
+__depends__ = {"20250617_01_c5mrF-task-split"}
+
+"""
+Yoyo migration to convert code_files table from TEXT to BYTEA
+"""
+
+
+def convert_code_to_bytea(conn):
+    """Convert existing TEXT code to BYTEA and recalculate hashes"""
```
b9r5, line 16:
```suggestion
    """Convert existing TEXT code to BYTEA"""
```

src/migrations/20250728_01_Q3jso-fix-code-table.py
```diff
@@ -0,0 +1,93 @@
...
+    # Get all existing records
+    cursor.execute("SELECT id, code FROM leaderboard.code_files")
+    records = cursor.fetchall()
+
+    for record_id, code_bytes in records:
+        code_text = bytes(code_bytes).decode("utf-8")
+        cursor.execute(
+            "UPDATE leaderboard.code_files SET old_code = %s WHERE id = %s",
+            (code_text.encode("utf-8"), record_id),
+        )
```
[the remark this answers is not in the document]
ngc92: the migration is run a a transaction, right? So that currently, if there's a UnicodeDecodeError we just wouldn't  change the DB at all.
AFAIK, all our submissions are valid unicode, though.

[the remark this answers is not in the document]
b9r5: Yes, it does run in a transaction, and if UnicodeError is raised here (or above at line 28) then the transaction would be rolled back.

We'd see the traceback and message, so I don't think there is any need to catch, log, and re-raise.

src/migrations/20250728_01_Q3jso-fix-code-table.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+steps = [
+    # prepare the table columns
+    step(
+        """
+        ALTER TABLE leaderboard.code_files DROP COLUMN hash;
+        ALTER TABLE leaderboard.code_files RENAME COLUMN code TO old_code;
+        ALTER TABLE leaderboard.code_files ADD COLUMN code BYTEA NOT NULL DEFAULT '';
```
b9r5, line 69: By the way, providing a default value when adding a column forces the acquisition of a table-level lock as the database rewrites the table to set the value of the new column. If the table is large, that lock can be held for a long time and will block concurrent reads and writes to the table. Not specifying a default value is typically a metadata-level change, and as such is almost instant. I don't think this matters in our database, which is fairly small, so that rewriting the table should be quite quick. When you test out the migration on a copy of prod, it would be a good idea to try running `alter table leaderboard.code_files add column foo bytea not null default ''` to see how long it takes.

src/migrations/20250728_01_Q3jso-fix-code-table.py
```diff
@@ -0,0 +1,94 @@
...
+        """,
+        """
+         ALTER TABLE leaderboard.code_files DROP COLUMN code;
+         ALTER TABLE leaderboard.code_files RENAME COLUMN old_code TO code;
+         ALTER TABLE leaderboard.code_files ADD COLUMN hash TEXT
+            GENERATED ALWAYS AS (encode(sha256(code::bytea), 'hex')) STORED;
+         """,
+    ),
+    # run the conversion
+    step(convert_code_to_bytea, convert_bytea_to_text),
```
b9r5, line 79: This step is taking a long time in my local setup. It might be worth adding a tqdm progress bar or even printing total number of records and something like a `.` or log line every 100 records or so. How long did this take in the dev environment? It's completely possible that it's taking a long time for me because I didn't give the container enough resources.

msaroufim (approved):

PR #334, opened 2025-08-22
Add submission api for web

yangw-dev:
## Description
Add an API serve for website gpumode, the api will return sub_id and sub_job_status_id immediately after the job is enqueue with background manager, then pull the data periodically for the update.

Adding queue mechnism to accept run jobs from api request /submission/ and runs the backend job.
the gpumode can then pull the db table to consistently check the status of the submission run.

### Why not fastapi background task?
https://fastapi.tiangolo.com/tutorial/background-tasks/

The fastapi background task is ok to use when the background job is <=30 secs such as sending emails, notificaiton. For us, it's not ideal since it uses same event loop. 

### Long term
Ideally we should handle submission jobs as background jobs using RQ or celery, but due to the time limit, this is a compromise solution. Some improvement can be done in next prs:
1. allow queue recoveries from the db data
2. handle shut down more gracefully, (for instance let running job finish, when heroku restarted) 

Or we should change the way we interact with the github runner, instead of keeping listening, we should provide a hook to let the workflow trigger the important update includes upload the result 

## Checklist

    ```bash
    python discord-bot.py
    ```
    ```
    ```
  README.md.


yangw-dev:
@S1ro1 
hMMM so i followed the guidane to setup my local env, but the /verilyruns still fail, not really sure how to verify that.
need some guidance.

but feel free to start the server and run this with the database:
```
INSERT INTO leaderboard.user_info (id,user_name, web_auth_id) VALUES ('123','alice', '234');
INSERT INTO leaderboard.gpu_type (leaderboard_id, gpu_type) VALUES (340, 'L4');
UPDATE leaderboard.leaderboard SET deadline = TIMESTAMPTZ '2026-09-01 23:59:59+00' WHERE id = 340;
```
the trigger the api:
```
 curl -N  -X POST "http://127.0.0.1:8000/submission/grayscale/L4/test" \
  -H "X-Web-Auth-Id: 234"
```

yangw-dev:
i setup this queue process due to the time limit, it's not perfect, but due to the time limit, i put it here:
in long term

In long term we should consider options:
1. passively let github call an api we provide for running update instead of keep listening 
or
2. introducte redis queue to handle the queue and background job properly

src/kernelbot/api/api_utils.py
```diff
@@ -183,3 +185,103 @@ async def display_report(self, title: str, report: RunResultReport):
             elif isinstance(part, Log):
                 self.long_report += f"\n\n## {part.header}:\n"
                 self.long_report += f"```\n{part.content}```"
+
+async def to_submit_info(
```
yangw-dev, line 189: i move the check logics to the function to handle things properly

src/libkernelbot/backend.py
```diff
@@ -53,20 +53,26 @@ def register_launcher(self, launcher: Launcher):
             self.launcher_map[gpu.value] = launcher
 
     async def submit_full(
-        self, req: ProcessedSubmissionRequest, mode: SubmissionMode, reporter: MultiProgressReporter
+        self, req: ProcessedSubmissionRequest, mode: SubmissionMode, reporter: MultiProgressReporter,
+        pre_sub_id: Optional[int] = None
```
yangw-dev, line 57: add this field since we want to receive the sub_id for the website api beforehand

src/kernelbot/api/main.py
```diff
@@ -316,94 +370,101 @@ async def run_submission(  # noqa: C901
...
-    )
+        # throw error if submission request is invalid
+        try:
+            req = prepare_submission(submission_request, backend_instance)
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=str(e)) from e
 
-    return StreamingResponse(generator, media_type="text/event-stream")
+        # prepare submission request before the submission is started
+        if not req.gpus or len(req.gpus) != 1:
```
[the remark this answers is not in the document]
yangw-dev: this is from prepare_submission can add it there

src/libkernelbot/backend.py
ngc92: does it make sense to turn this function into a generator and yield the submission id once it is created?

S1ro1 (approved):
Not 100% happy with the complexity it adds, but if it works we can always fix it later (we never do). Tested locally with CLI, works just fine

PR #335, opened 2025-08-23
Multi-GPU

ngc92:
Implement multi-gpu tasks:
design: 
run_eval still calls only a single instance of eval.py, but if the task is multi-gpu, eval.py will create a pool of multiple worker processes, and call the user op on each with a different rank argument.

in the current code, input generation and testing is also handled by each process independently; I'm not sure if that's the right approach.

### DIFF
--- a/examples/eval.py
+++ b/examples/eval.py
@@ -197,7 +197,8 @@ def run_multi_gpu_test(pool: multiprocessing.Pool, test: TestCase, world_size: i
                 args=(test, i),
             )
         )
-    rets = [el.get() for el in rets]
+    # 60 seconds should be more than enough, we want tests to be fast
+    rets = [el.get(60) for el in rets]
 
     correct = all(ret[0] for ret in rets)
     error_messages = str.join("\n", [f"rank {rank} - {ret[1]}" for rank, ret in enumerate(rets) if not ret[0]])
@@ -422,7 +423,9 @@ def run_multi_gpu_benchmark(pool: multiprocessing.Pool, test: TestCase, recheck:
                 args=(test, i, recheck, max_repeats, max_time_ns),
             )
         )
-    rets = [el.get() for el in rets]
+
+    # 120 seconds for benchmarking + we run a pre-benchmark test and want to leave some slack
+    rets = [el.get(timeout=180) for el in rets]
 
     # For multi-GPU benchmarking, only rank 0 has meaningful stats
     failed_ranks = []
--- a/examples/eval.py
+++ b/examples/eval.py
@@ -211,7 +211,7 @@ def run_single_test(pool: multiprocessing.Pool, test: TestCase):
     """
     world_size = test.args.get("world_size", None)
     if world_size is None:
-        return pool.apply(_run_single_test, (test, 0, 0))
+        return pool.apply(_run_single_test, (test,))
     else:
         return run_multi_gpu_test(pool, test, world_size)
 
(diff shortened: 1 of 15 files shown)


examples/eval.py
```diff
@@ -157,16 +159,60 @@ def _run_single_test(test: TestCase):
...
 
+def _run_distributed_test(test: TestCase, rank: int):
+    """
+    Runs a single test case. Do not call directly
+    """
+    from submission import custom_kernel
+    import torch.distributed as dist
+    world_size = test.args["world_size"]
+    os.environ["MASTER_ADDR"] = "127.0.0.1"
+    os.environ["MASTER_PORT"] = "12356"
```
msaroufim, line 175: might wanna setup a large random port, if for some reason a job fails and the port doesn't get released

examples/eval.py
```diff
@@ -261,6 +307,142 @@ def _run_single_benchmark(test: TestCase, recheck: bool, max_repeats: int, max_t
...
+    dist.init_process_group("nccl", init_method="env://", rank=rank, world_size=world_size, device_id=torch.device(f'cuda:{rank}'))
+
+    try:
+        durations = []
+        # generate input data once
+        data = generate_input(**test.args, rank=rank)
+        check_copy = _clone_data(data, rank)
+
+        # first, one obligatory correctness check
+        output = custom_kernel(_clone_data(data, rank))
```
danielhua23, line 330: I'd like to confirm, if that, users' custom_kernel should be designed to an API accepting single rank data? but my ref_kernel is accepting all rank data...is that conflicted? https://github.com/gpu-mode/reference-kernels/pull/51/files#diff-4634bd7a4a47ab89859ee0db3f4f3f3c8123cf18981fb4d24b4655a412777013R240

ngc92:
> This is a really cool PR, I don't think I've seen anyone recently manually manage process pools as opposed to just calling torchrun. I think this solution works quite nicely for us though because we don't care (yet) about multi node or fault tolerance
> 
> Idk if he has time but just in case @kiukchung would really appreciate your feedback on whether this is a reasonable way of supporting a job system where people submit distributed kernels
> 
> The part I'm a bit worried about is the example right now can check that a kernel is globally correct by iterating over many ranks and making sure they're locally correct. That's probably not going to be true for any of the distributed AMD kernels. This is fine as long as we offload the burden of gathering all the outputs onto the user

my first attempt was to just `torchrun eval.py`, but that breaks our file descriptor approach for result reporting.  so i followed @danielhua23's example and did the process startup manually,  which fit quite nicely with the existing code that had a process pool of one process anyway.

If you cannot do the check locally, there's nothing preventing us from having all-gather calls inside the validation function.

examples/eval.py
```diff
@@ -157,16 +159,60 @@ def _run_single_test(test: TestCase):
...
+    rets = []
+    # world_size is a mandatory argument for multi-gpu tests
+    for i in range(world_size):
+        rets.append(
+            pool.apply_async(
+                _run_distributed_test,
+                args=(test, i),
+            )
+        )
+    rets = [el.get() for el in rets]
```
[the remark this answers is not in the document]
ngc92: done

src/libkernelbot/backend.py
```diff
@@ -29,13 +29,8 @@ def __init__(
     ):
         self.debug_mode = debug_mode
         self.db = LeaderboardDB(
-            env.POSTGRES_HOST,
```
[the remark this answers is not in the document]
ngc92: rebased

examples/eval.py
ngc92: I think essentially, the user kernel should be what is called `_worker` in your code. I wanted to avoid passing tensor objects across process boundaries with the multiprocessing module; I see in your example, you also transfer the tensor to the CPU before returning from worker.

danielhua23: yes, `_worker` is actually the kernel for each rank to run. but I am worried that people will not regard `_worker` as `custom_kernel`, but `ref_kernel` as `custom_kernel`, then Can your code successfully handle the case?

if get people regard `_worker` as `custom_kernel`, should I remove the current `ref_kernel` and change `_worker` to `ref_kernel`?

msaroufim (approved):

PR #339, opened 2025-08-25
Initial AMD Profiling

Snektron:
## Description

This is an initial implementation of profiling using rocPROF for MI300 jobs. It works as follows: When the profiling run time is selected, the application is simply executed using rocPROF with options set to gather a bunch of traces. These are then packages with RunResults, and finally sent to the user. This current version is the most straight forward way to implement this, but there are some missing features and limitations:
- [x] The tracing results ZIP needs to be less than 8 MB. While Discord should support 25 MB for non-boosted servers and 100 MB with boosted servers, this is apparently an API limitation. See https://github.com/Rapptz/discord.py/issues/1733. Test traces appear to be about 5 MB, but I don't know yet about longer programs and multi-GPU programs.
  - A github link is now shared
- [ ] The HSA trace is not collected. It is too large to even be opened by perfetto in a reasonable time, so for now I've left that out. Perhaps later.
  - I think its best to leave this out
- [x] I've not yet tested how well it works with multi-GPU programs. I'll check by rebasing on top of https://github.com/gpu-mode/discord-cluster-manager/pull/335.
- [x] I'd like to add code object dumps to the profile results, but those increase the binary by quite a bit too.
- [x] `eval.py` does not yet contain rocTX markers.
- [ ] rocPROF prints a bunch of perfetto messages which always appear in the stderr. This looks like an issue on rocPROF's side, logging isn't configured properly. Not sure if we should solve this now...

### DIFF
--- a/src/libkernelbot/run_eval.py
+++ b/src/libkernelbot/run_eval.py
@@ -4,6 +4,7 @@ import functools
 import json
 import os
 import shlex
+import shutil
 import subprocess
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


msaroufim:
Sorry for the delay, I'll have time to review tomorrow afternoon PST

ngc92:
quickly scrolled over the changes, didn't spot anything that looked problematic. I'd echo Mark's comment, though, that it'd be nice if the PR was a bit smaller :)
There's a bunch of changes here that aren't strictly about profiling (the example updates, but also, e.g., reporting cuda vs rocm); if we could get them in separate PRs, those would be trivial to approve and this would be a bit more focused.

PR #354, opened 2025-09-09
Profiling Infrastructure

Snektron:
## Description

This PR adds some infrastructure for dealing with profiling data. Basically, the idea is that any data which is placed in the `profile_data/` directory in the github runner is exported (when `EvalResult.profile_result` is set) to the user via discord. As discord attachments could not handle the size of download artifacts (typically up to ~35 MB), I've opted to simply present a direct link to the GitHub artifact. To this end, I've modified the GH launcher to provide a sort of 'index' of artifacts, which can then either be downloaded by the bot or presented as download link.

I've also fixed some minor bugs in `run_eval.py` related to fetching ROCm system info, as well as added some extra info to `SystemInfo` about the runtime (useful elsewhere in the evaluating process when the actual profiling stuff is added).

Some caveats:
* Users can export any file as download by profiling and writing to `profile_data/` currently. I think that the only way around that would be to launch the profiling process with higher privileges and to then drop those privileges in `eval.py`. Lmk if you guys want that.
* Profiling artifacts aren't protected, they can be downloaded by anyone (provided that they have a GH account), possibly giving away information about a solution. It shouldn't contain kernel source, just kernel names + run order.

Extracted from #339

### DIFF
--- a/.github/workflows/amd_workflow.yml
+++ b/.github/workflows/amd_workflow.yml
@@ -35,13 +35,13 @@ jobs:
       run: |
         # Extract the payload content without printing it
         PAYLOAD=$(jq -r '.inputs.payload' $GITHUB_EVENT_PATH)
-        
+
         # Apply mask to the extracted content
         echo "::add-mask::$PAYLOAD"
-        
+
         # Now write to file (won't be logged since it's masked)
         echo "$PAYLOAD" > payload.json
-    
+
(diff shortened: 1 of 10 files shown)


msaroufim:
Current caveats seem fine to me, the retention policy makes it so you don't have too much time to be abusive. Feel free to merge

PR #408, opened 2026-02-01
Add API-only mode and admin CLI support

msaroufim:
This PR needs to be reviewed and tested in conjunction with https://github.com/gpu-mode/popcorn-cli/pull/29

It's now possible to pass an `--api-only` flag to start the FastAPI server without Discord credentials. Which means admins are now determined by a hardcoded token, locally you can set this to whatever but we'll have to hardcode a specific one in heroku

I also added a skills.md for future AI so they can now fully iterate and manage e2e bot tests fully in the terminal

Also more of a drive by contribution but I changed the default behavior of not setting a gpu in a leaderboard to error, since default to H100 just seemed confusing to me

This PR is quite short, it's only the updates to md files for the AIs that are long, I don't expect a thorough review there

Overall this has been a substantial qol improvement for me because I just pass AIs a concrete task and they can finally test everything end to end without clicking buttons in discord

### DIFF
--- a/tests/test_admin_api.py
+++ b/tests/test_admin_api.py
@@ -259,3 +259,149 @@ class TestAdminLeaderboards:
         assert response.status_code == 200
         assert response.json()["force"] is True
         mock_backend.db.delete_leaderboard.assert_called_once_with("test-leaderboard", force=True)
+
+
+class TestAdminUpdateProblems:
+    """Test admin update-problems endpoint."""
+
+    def test_update_problems_requires_auth(self, test_client):
+        """POST /admin/update-problems requires authorization."""
+        response = test_client.post("/admin/update-problems", json={})
+        assert response.status_code == 401
+
+    def test_update_problems_success(self, test_client, mock_backend):
+        """POST /admin/update-problems returns sync results."""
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Local Development
+
+See `SKILLS/test_bot.md` for local testing setup instructions.
+
+## Architecture
+
+### Problem Configuration
+
+Problems are defined in two ways:
```
ngc92: point at our examples directory

src/kernelbot/api/main.py
```diff
@@ -470,6 +486,123 @@ async def run_submission_async(
...
+@app.post("/admin/stop")
+async def admin_stop(
+    _: Annotated[None, Depends(require_admin)],
+) -> dict:
+    backend_instance.accepts_jobs = False
+    return {"status": "ok", "accepts_jobs": False}
+
+
+@app.post("/admin/leaderboards")
+async def admin_create_leaderboard(
```
ngc92: should be `create_dev_leaderboard`?

S1ro1 (approved):
lgtm

PR #412, opened 2026-02-01
Add user submission management API endpoints

msaroufim:
Should be reviewed in conjunction with https://github.com/gpu-mode/popcorn-cli/pull/32

- Add get_user_submissions() DB method to query user's submissions
- Add GET /user/submissions endpoint to list user's submissions
- Add GET /user/submissions/{id} endpoint to view submission with code
- Add DELETE /user/submissions/{id} endpoint to delete own submissions
- All endpoints verify ownership before allowing access

Fixes https://github.com/gpu-mode/kernelbot/issues/364

### DIFF
--- a/src/kernelbot/api/main.py
+++ b/src/kernelbot/api/main.py
@@ -777,6 +777,8 @@ async def get_user_submission(
             if str(submission["user_id"]) != str(user_info["user_id"]):
                 raise HTTPException(status_code=403, detail="Not authorized to view this submission")
 
+            # RunItem is a TypedDict (already a dict), select fields to expose
+            run_fields = ("start_time", "end_time", "mode", "secret", "runner", "score", "passed")
             return {
                 "id": submission["submission_id"],
                 "leaderboard_id": submission["leaderboard_id"],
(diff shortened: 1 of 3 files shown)


src/kernelbot/api/main.py
```diff
@@ -711,3 +711,130 @@ async def get_submission_count(
...
+                        "start_time": r["start_time"],
+                        "end_time": r["end_time"],
+                        "mode": r["mode"],
+                        "secret": r["secret"],
+                        "runner": r["runner"],
+                        "score": r["score"],
+                        "passed": r["passed"],
+                    }
+                    for r in submission["runs"]
+                ],
```
ngc92: to_dict utility function?

src/libkernelbot/leaderboard_db.py
```diff
@@ -848,6 +848,86 @@ def delete_submission(self, submission_id: int):
...
+                        s.file_name,
+                        s.submission_time,
+                        s.done,
+                        r.runner as gpu_type,
+                        r.score
+                    FROM leaderboard.submission s
+                    JOIN leaderboard.leaderboard lb ON s.leaderboard_id = lb.id
+                    LEFT JOIN leaderboard.runs r ON r.submission_id = s.id
+                        AND NOT r.secret AND r.passed
+                    WHERE s.user_id = %s
```
ngc92: unify with the string above, the only difference is  the WHERE part?

src/libkernelbot/leaderboard_db.py
```diff
@@ -848,6 +848,82 @@ def delete_submission(self, submission_id: int):
...
+                    ORDER BY s.submission_time DESC
+                    LIMIT %s OFFSET %s
+                """
```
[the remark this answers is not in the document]
ngc92: I  think this is wrong, we  shouldn't just fetch one run of a given submission, that would be unexpected. Instead, we probably want the code below to restructure result to be a dict, with an item "runs" that is a list of (gpu_type, score)  tuples?

PR #509, opened 2026-07-05
Add CUTLASS DSL to Helion runner image

nataliakokoromyti:
## Summary
- install `nvidia-cutlass-dsl==4.5.2`, `cuda-core[cu13]`, and `cuda-python[all]==13.0` in the Helion runner image
- install CUTLASS v4.5.1 C++ headers under `/opt/cutlass`
- export `CUTLASS_PATH` and `CPLUS_INCLUDE_PATH` so CUDA extension builds can find CUTLASS headers

## Why
The hosted B200/Brev profiling runtime currently cannot run CuTe DSL submissions because `import cutlass`, `import cuda`, and `import cuda.bindings` fail with `ModuleNotFoundError`. The Modal runner already carries this dependency stack; this enables the same CuTe/CUTLASS path for the Helion runner image.

## Validation
- `git diff --check`
- resolved the added package set with `uv pip compile` for Python 3.13 on `x86_64-manylinux_2_28` using binary wheels only

### DIFF
--- a/docker/helion.Dockerfile
+++ b/docker/helion.Dockerfile
@@ -83,14 +83,13 @@ RUN sudo uv pip install --system \
 # # tinygrad
 # RUN sudo uv pip install --system tinygrad~=0.10
 
-# # NVIDIA CUDA packages
-# RUN sudo uv pip install --system \
-#     nvidia-cupynumeric~=25.3 \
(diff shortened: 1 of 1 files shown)


msaroufim:
Not 100% sure but I think you might have meant to make a PR here? https://github.com/gpu-mode/ncu-service

